PR #1, opened 2021-04-20
Initial setup and DW jobtap plugin

SteVwonder:
Copies the initial skeleton build/test system from flux-core/sched and then adds the jobtap plugin for DW string validation


t/sharness.d/flux-sharness.sh
```diff
@@ -0,0 +1,244 @@
+
```
grondo, line 1: This file needs to be updated from `flux-core`. (e.g. the `--bootstrap` option was removed from `flux-start` recently, and the `--size` option name changed.)

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,122 @@
...
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* dws-test.c - keep jobs in depend state if they contain a dws attribute,
```
grondo, line 11: `s/dws-test.c/dws-jobtap.c/`?

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,122 @@
...
+
+static void validate_cb (flux_future_t *f, void *arg)
+{
+    struct validate_arg_t *args = arg;
+    int success = false;
+    char *errstr = NULL;
+    flux_t *h = flux_future_get_flux(f);
+    char jobid_buf[64];
+
+    flux_job_id_encode (args->id, "f58", jobid_buf, 64);
```
grondo, line 40: If this is meant to simply be a prototype, feel free to ignore this comment, but there are several places missing return code checks. This could be fixed in a future update to "non-prototype" as far as I'm concerned.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,122 @@
...
+    if (flux_rpc_get_unpack (f,
+        "{s:b, s?s}",
+        "success", &success,
+        "errstr", &errstr) < 0)
```
grondo, line 44: We've unofficially adopted a style of aligning function arguments on newlines with the first argument, e.g.:
```suggestion
    if (flux_rpc_get_unpack (f,
                             "{s:b, s?s}",
                             "success", &success,
                             "errstr", &errstr) < 0)
```

Optional, since this is a separate repo

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,122 @@
...
+        char *reason;
+        if (asprintf (&reason, "DW string validation failed: %s", errstr) < 0)
+            goto done;
+        flux_job_raise (h, args->id, "dw-validation", 0, reason);
+
+        goto done;
+    }
+
+done:
+    free (args);
```
grondo, line 64: The `flux_future_t` needs to be destroyed here as well. The `args` could be added to the future with `flux_future_aux_set()` to tie destruction of `args` to the `flux_future_t`.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,122 @@
...
+    if (success) {
+        flux_jobtap_dependency_remove (args->p, args->id, VALIDATE_DEP_NAME);
+    } else {
+        flux_log_error (h, "Failed to validate DW string for job %s", jobid_buf);
+        
+        char *reason;
+        if (asprintf (&reason, "DW string validation failed: %s", errstr) < 0)
+            goto done;
+        flux_job_raise (h, args->id, "dw-validation", 0, reason);
+
```
grondo, line 59: `reason` is leaked here.

grondo (commented):
I added a few inline comments before I realized this is probably a prototype and doesn't need a full code review.

The basic plugin design looks fine to me! There are some leaks, unchecked return codes, and (optional) whitespace fixups that might need to be addressed before this plugin would be considered "production" ready.

Nice work on the repo!

grondo:
If you also copy `t/valgrind/valgrind.supp` into place from `flux-core` then you will be able to use `FLUX_TEST_VALGRIND=t` on sharness tests to check for leaks (when the tests here get more advanced that may be handy)

(This makes me wonder if there is a way for flux-core to "install" some of this sharness infrastructure in a better way so it doesn't have to be copied into framework projects..)

SteVwonder:
Thanks @grondo for the thorough review!  I believe I addressed all of your feedback, and I enabled valgrind in the CI.  Can you quick the two new fixup commits a quick look?  I'm going to open a separate issue about refactoring to take advantage of the new scheme parsing in flux-core.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -0,0 +1,149 @@
...
+        flux_future_t *validate_fut = flux_rpc_pack (
+            h, "dws.validate", FLUX_NODEID_ANY, 0, "{s:s}", "dw_string", dw
+        );
+        if (validate_fut == NULL) {
+            flux_log_error (h, "Failed to send dws.validate RPC");
+            rc = -1;
+            goto ret;
+        }
+        struct validate_arg_t *validate_args = calloc (sizeof (struct validate_arg_t), 1);
```
grondo, line 118: check for `calloc()` failure.

grondo (approved):
LGTM! Just one missing check for memory allocation failure.
Feel free to fix that and squash the result

SteVwonder:
Thanks @grondo.  Added the calloc error check and squashed. Merging

PR #7, opened 2021-08-25
Shell: libpals interface plugins

### LINKED ISSUE #6: Cray PMI (libpals) support
In order for Flux to support Cray's PMI on Shasta machines like rznevada, Perlmutter, and El Cap, Flux will need to interface with Cray's libpals. From the Cray folks:

> To support Cray PMI, the launcher must perform these tasks:
> 1. Assign an apid to the application (arbitrary string, unique per-application)
> 2. On each compute node, create a spool directory owned by the application's user
> 3. On each compute node, write an apinfo file in the spool directory (format below)
> 4. Set environment variables for each spawned process (listed below)
> 5. Remove the spool directory on each compute node when the application is complete
>  
> The apinfo file uses C structures written directly to the file. It starts with a header, followed by a number of structure arrays.
>  
> These environment variables should be set for each process:
> PALS_APID - Application ID (arbitrary string, mostly used for logging)
> PALS_APINFO - Full path to the apinfo file
> PALS_RANKID - Global rank ID for this process
> PALS_NODEID - Node index for this process (e.g. head compute node is 0, next compute node is 1, etc)
> PALS_SPOOL_DIR - Application-specific directory for keeping runtime files
> PMI_CONTROL_PORT - Port number for PMI to bind on each compute node. Must be different for each concurrent application running on the same node(s).

jameshcorbett:
Moved from [flux-core#3826](https://github.com/flux-framework/flux-core/pull/3826), closes #6.

### DIFF
--- a/src/job-manager/plugins/Makefile.am
+++ b/src/job-manager/plugins/Makefile.am
@@ -7,7 +7,7 @@ AM_LDFLAGS = \
 	$(CODE_COVERAGE_LIBS)
 
 jobtapdir = \
-    $(fluxlibdir)/job-manager/
+    $(fluxlibdir)/job-manager/plugins/
 
 jobtap_LTLIBRARIES = \
 	dws-jobtap.la \
--- a/src/job-manager/plugins/Makefile.am
+++ b/src/job-manager/plugins/Makefile.am
@@ -9,7 +9,9 @@ AM_LDFLAGS = \
 jobtapdir = \
     $(fluxlibdir)/job-manager/
 
-jobtap_LTLIBRARIES = dws-jobtap.la
+jobtap_LTLIBRARIES = \
+	dws-jobtap.la \
+	cray_pals_port_distributor.la
 
 dws_jobtap_la_SOURCES = \
 	dws-jobtap.c
@@ -22,3 +24,14 @@ dws_jobtap_la_LDFLAGS = \
 dws_jobtap_la_LIBADD= \
     $(FLUX_CORE_LDFLAGS)
 
+cray_pals_port_distributor_la_SOURCES = \
+	cray_pals_port_distributor.c
+
+cray_pals_port_distributor_la_LDFLAGS = \
+	$(AM_LDFLAGS) \
+	-module \
+	-avoid-version
+
+cray_pals_port_distributor_la_LIBADD= \
+    $(FLUX_CORE_LIBS) \
+    $(FLUX_HOSTLIST_LIBS)
--- a/src/job-manager/plugins/cray_pals_port_distributor.c
+++ b/src/job-manager/plugins/cray_pals_port_distributor.c
@@ -1,0 +1,281 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* cray_pals_port_distributor.c - Distribute port numbers for use
+ * by Cray's libpals. See also `src/shell/plugins/cray_pals.c`.
+ */
+
+#include <stdio.h>
+#include <syslog.h>
+
+#include <jansson.h>
+
+#include <flux/core.h>
+#include <flux/hostlist.h>
+#include <flux/jobtap.h>
+
+#define CRAY_PALS_AUX_NAME "cray::libpals::ports"
+
+struct port_range {
+    int *available_ports;
+    int fill;
+    int maxsize;
+};
+
+struct plugin_range_jobid_triple {
+    flux_plugin_t *plugin;
+    struct port_range *range;
+    flux_jobid_t jobid;
+};  // for passing to flux_future_t callback
+
+static int get_port (struct port_range *range)
+{
+    if (range->fill <= 0) {
+        return -1;
+    }
+    range->fill--;
+    return range->available_ports[range->fill];
+}
+
+static int set_port (struct port_range *range, int port)
+{
+    if (range->fill >= range->maxsize) {
+        return -1;
+    }
+    range->available_ports[range->fill] = port;
+    range->fill++;
+    return 0;
+}
+
+/*
+ * Return a 'struct hostlist' containing the hostnames of every shell rank.
+ */
+static struct hostlist *hostlist_from_array (json_t *nodelist_array)
+{
+    size_t index;
+    json_t *value;
+    struct hostlist *hlist;
+    const char *entry;
+
+    if (!json_is_array (nodelist_array) || !(hlist = hostlist_create ())) {
+        return NULL;
+    }
+    json_array_foreach (nodelist_array, index, value)
+    {
+        if (!(entry = json_string_value (value))
+            || hostlist_append (hlist, entry) < 0) {
+            hostlist_destroy (hlist);
+            return NULL;
+        }
+    }
+    return hlist;
+}
+
+/* Callback to KVS lookup for a job's R.
+ * Calculate the number of shells in the job and then optionally
+ * post a cray_port_distribution to the job's eventlog.
+ */
+static void count_job_shells (flux_future_t *fut, void *arg)
+{
+    struct plugin_range_jobid_triple *triple = arg;
+    json_t *nodelist;
+    struct hostlist *hlist = NULL;
+    int port1, port2;
+    flux_t *h;
+    json_t *arr = NULL;
+
+    if (!(h = flux_future_get_flux (fut))) {
+        goto cleanup;
+    }
+    if (flux_kvs_lookup_get_unpack (fut,
+                                    "{s:{s:o}}",
+                                    "execution",
+                                    "nodelist",
+                                    &nodelist) < 0
+        || !(hlist = hostlist_from_array (nodelist))) {
+        flux_log_error (h,
+                        "cray_pals_port_distributor: "
+                        "Error fetching R from shell-counting future");
+        goto cleanup;
+    }
+    if (hostlist_count (hlist) == 1) {
+        goto cleanup;  // no need to post ports
+    }
+    // assign ports to the job
+    if ((port1 = get_port (triple->range)) < 0 || (port2 = get_port (triple->range)) < 0
+        || !(arr = json_pack ("[i, i]", port1, port2))
+        || flux_jobtap_event_post_pack (triple->plugin,
+                                        triple->jobid,
+                                        "cray_port_distribution",
+                                        "{s:O}",
+                                        "ports",
+                                        arr) < 0
+        || flux_jobtap_job_aux_set (triple->plugin,
+                                    triple->jobid,
+                                    CRAY_PALS_AUX_NAME,
+                                    arr,
+                                    (flux_free_f)json_decref) < 0) {
+        if (arr)
+            json_decref (arr);
+        flux_log_error (h,
+                        "cray_pals_port_distributor: "
+                        "Failed to post ports to job");
+        goto cleanup;
+    }
+
+cleanup:
+    if (hlist)
+        hostlist_destroy (hlist);
+    flux_future_destroy (fut);
+    free (triple);
+    return;
+}
+
+/* Create a future to get a job's R. Callback to the future will grab
+ * free ports and write them into the job's event log.
+ */
+static int run_cb (flux_plugin_t *p,
+                   const char *topic,
+                   flux_plugin_arg_t *args,
+                   void *arg)
+{
+    flux_t *h;
+    flux_jobid_t jobid;
+    char buf[1024];
+    flux_future_t *fut = NULL;
+    struct plugin_range_jobid_triple *triple;
+    // 'arg' is 'struct port_range *'
+
+    if (!(h = flux_jobtap_get_flux (p))
+        || flux_plugin_arg_unpack (args, FLUX_PLUGIN_ARG_IN, "{s:I}", "id", &jobid) < 0
+        || !(triple = malloc (sizeof (struct plugin_range_jobid_triple))))
+        return -1;
+    triple->range = arg;
+    triple->jobid = jobid;
+    triple->plugin = p;
+    if (flux_job_kvs_key (buf, sizeof (buf), jobid, "R") < 0
+        || !(fut = flux_kvs_lookup (h, NULL, 0, buf))
+        || flux_future_then (fut, -1., count_job_shells, triple) < 0) {
+        if (fut)
+            flux_future_destroy (fut);
+        flux_log_error (h,
+                        "cray_pals_port_distributor: "
+                        "Error creating shell-counting future");
+        free (triple);
+        return -1;
+    }
+    // 'triple' freed in callback
+    return 0;
+}
+
+/* On a job's cleanup event, get the ports and return them
+ * to the pool.
+ */
+static int cleanup_cb (flux_plugin_t *p,
+                       const char *topic,
+                       flux_plugin_arg_t *args,
+                       void *arg)
+{
+    struct port_range *range = arg;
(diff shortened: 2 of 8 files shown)


jameshcorbett:
Removing WIP, I think this is ready to review with the exception of tests. I gave it a few runs on RZNevada and I was able to run 4000 MPI jobs with no failures.

There is currently no special logic for handing out ports to nested instances, BUT if the nested instance:
a) Has exclusive access to its nodes
OR b) fits on a single node

then nothing needs to be done. So only instances that spread over multiple nodes with non-exclusive access would be a problem. But I think that’s probably pretty rare, I’m not sure why anyone would want to do that, so I think that’s good enough for now

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -0,0 +1,243 @@
...
+#include <jansson.h>
+#include <stdio.h>
+
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "../../shell/plugins/eventlog.h"
+
+#define CRAY_PALS_PORT_MIN 10000
+#define CRAY_PALS_PORT_MAX 12000
```
jameshcorbett, line 28: This works for now but does anyone have a better idea of how this could be configured at build or run time? One option would be to look in `/etc/slurm/slurm.conf` for `MPIParams=ports=N-M` and  just use that (on RZNevada that comes out to 11000-12000).

src/shell/plugins/Makefile.am
```diff
@@ -0,0 +1,24 @@
...
+    $(fluxlibdir)/shell/plugins/
+
+shell_plugin_LTLIBRARIES = cray_pals.la
+
+cray_pals_la_SOURCES = cray_pals.c
+cray_pals_la_CPPFLAGS = $(AM_CPPFLAGS)
+cray_pals_la_LIBADD = \
+	$(FLUX_CORE_LDFLAGS) \
+	$(fluxlibdir)/../libflux-hostlist.la \
+	$(fluxlibdir)/../libflux-core.la \
```
jameshcorbett, line 19: Is there a better way to do this? Seems a little hacky

src/shell/plugins/eventlog.h
```diff
@@ -0,0 +1,129 @@
+#include <jansson.h>
```
jameshcorbett, line 1: With the exception of the last function, these functions are copied over from flux-core because they aren't currently exported. I will see about fixing that.

src/job-manager/plugins/cray_pals_port_distributor.c
grondo: There is a facility for adding configuration to jobtap plugins, see the [CONFIGURATION](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/man7/flux-jobtap-plugins.html#configuration) section of the manpage.

You can access the configuration for the plugin with `flux_plugin_get_conf(3)` or more likely `flux_plugin_conf_unpack(3)`

src/job-manager/plugins/Makefile.am
```diff
@@ -22,3 +24,15 @@ dws_jobtap_la_LDFLAGS = \
...
+
+cray_pals_port_distributor_la_LDFLAGS = \
+	$(AM_LDFLAGS) \
+	-module \
+	-avoid-version
+
+cray_pals_port_distributor_la_LIBADD= \
+    $(FLUX_CORE_LDFLAGS) \
+    $(fluxlibdir)/../libflux-hostlist.la \
+	$(fluxlibdir)/../libflux-core.la
```
grondo, line 38: You should be able to build plugins against the installed flux-core using `-lflux-hostlist -lflux-core`. If that wasn't working for some reason we should determine if there's a bug in the flux-core install.

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -0,0 +1,243 @@
...
+#include "config.h"
+#endif
+
+#include <jansson.h>
+#include <stdio.h>
+
+#include <flux/core.h>
+#include <flux/jobtap.h>
+
+#include "../../shell/plugins/eventlog.h"
```
grondo, line 25: This might break when doing a vpath build. Typically autotools projects include the top level srcdir in their include PATH by default, so try `#include "src/shell/plugins/eventlog.h"`?

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -0,0 +1,243 @@
...
+        || flux_plugin_arg_unpack (args, FLUX_PLUGIN_ARG_IN, "{s:I}", "id", &jobid) < 0
+        || !(triple = malloc (sizeof (struct plugin_range_jobid_triple))))
+        return -1;
+    triple->range = arg;
+    triple->jobid = jobid;
+    triple->plugin = p;
+    if (flux_job_kvs_key (buf, sizeof buf, jobid, "R") < 0
+        || !(fut = flux_kvs_lookup (h, NULL, 0, buf))
+        || flux_future_then (fut, -1., count_job_nnodes, triple) < 0) {
+        flux_log_error (h, "Error creating node-counting future");
```
grondo, line 139: Is `fut` leaked on the error path here?

src/shell/plugins/Makefile.am
garlick: This might be what you're looking for?

https://github.com/flux-framework/flux-sched/blob/master/config/ax_flux_core.m4

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -0,0 +1,243 @@
...
+                       void *arg)
+{
+    flux_jobid_t id;
+    flux_t *h;
+    flux_future_t *fut;
+    // 'arg' is 'struct port_range *'
+
+    if (!(h = flux_jobtap_get_flux (p))
+        || flux_plugin_arg_unpack (args, FLUX_PLUGIN_ARG_IN, "{s:I}", "id", &id) < 0
+        || !(fut = flux_job_event_watch (h, id, "eventlog", 0))
```
grondo, line 209: I apologize if I'm missing something obvious, but I don't think you should have to replay the eventlog before freeing the ports associated with a job. Instead, when ports are allocated to a job, you could stash the list of ports as an "aux" item associated with the jobid, e.g.

```
   flux_jobtap_job_aux_set (p, JOBID, "cray::ports",  triple, free_triple);
```

(For example if you were to use the `triple` object for this)

Then, here in the job's `cleanup_cb()`, you can check for that aux item associated with the job using `flux_jobtap_job_aux_get(3)`. If it doesn't exist there is nothing to do. If it does, then you proceed to release the ports. No need to free `triple`, that will automatically be taken care of when the job becomes inactive.

It may not make sense to use the `struct plugin_range_jobid_triple` as the object to associate with the job (I haven't studied the code long enough to know if that is the case), but I just used that structure as an example.

grondo (commented):
Really nice work and congrats on getting it all working!

I only took a quick skim through the jobtap plugin and had some suggestions, so I thought I'd post them now.

Sorry you had to duplicate the eventlog convenience code here.. I wonder if we can offer an interface in the flux API to help with that. (We can't export functions that use `json_t` so it might be some small amount of development to get there. Maybe for another time)

src/job-manager/plugins/cray_pals_port_distributor.c
jameshcorbett: Good catch thanks!

src/job-manager/plugins/Makefile.am
jameshcorbett: That works!

src/job-manager/plugins/cray_pals_port_distributor.c
jameshcorbett: Works great and the code is much simpler! I ended up using a `json_array` instead of `triple`.

jameshcorbett:
Thanks for the great comments @grondo (and all the help on slack too). I believe I have addressed all of them now.

jameshcorbett:
Any advice on how to test given that the jobtap plugin doesn't really do anything unless a job takes > 1 nodes? I could change it to make it > 1 shells instead so that it could be tested with `-s2`, without much loss of optimization.

Also, is there a good way to make an instance pass its broker config down to its children, other than the `FLUX_CONFIG_DIR` environment variable? On Shasta systems every instance should have the jobtap plugin loaded with some range of ports. Just wondering if there is an alternative to making users define a small TOML file and point to it with the environment variable.

garlick:
> Any advice on how to test given that the jobtap plugin doesn't really do anything unless a job takes > 1 nodes? I could change it to make it > 1 shells instead so that it could be tested with -s2, without much loss of optimization.

Yeah, that's what we do in flux-core.  Effectively `flux start -sN` where N > 1 is like N nodes.  For most cases, that works, although the info that two brokers are really on one node leaks through in the process mapping.

> Also, is there a good way to make an instance pass its broker config down to its children, other than the FLUX_CONFIG_DIR environment variable?

The broker `--config-path` option is another way to specify the config file, in case that helps (probably not).  There isn't any inheritance support for configs.  Maybe the KVS could be used?  E.g. in the enclosing instance, write the allocated port range to the job directory (like _R_) and then in the new instance, have the jobtap plugin read it out using the job-info service?  (Hint: see [resource module example](https://github.com/flux-framework/flux-core/blob/67999f2af7c5374124d4c8933249204deb93c2b2/src/modules/resource/inventory.c#L397)).

grondo:
>  I could change it to make it > 1 shells instead so that it could be tested with -s2, without much loss of optimization.

I think if you just remove the call to `hostlist_uniq()` then you can test under `flux start -s N` as @garlick suggested.

> and then in the new instance, have the jobtap plugin read it out using the job-info service? (Hint: see resource module example).

Unfortunately, the problem is that this jobtap plugin won't be automatically loaded in the new instance. 

> Also, is there a good way to make an instance pass its broker config down to its children, other than the FLUX_CONFIG_DIR environment variable? On Shasta systems every instance should have the jobtap plugin loaded with some range of ports. Just wondering if there is an alternative to making users define a small TOML file and point to it with the environment variable.

I'm not sure why `FLUX_CONFIG_DIR` wouldn't work in this case. We can set it to `/etc/flux/shasta` or something in the default shell `initrc.lua`, so that it is only set for Flux jobs, and not in the system instance.
(Also not for `flux start -s N ` test instances)

grondo (commented):
Just one minor comment after a quick skim. Looks like really good work to me!

I'll take another pass once tests have been added. Thanks for this work!

jameshcorbett:
Tests added, many thanks to @grondo for lots of help with debugging the tests

grondo:
The tests appear to be failing because the `fluxrm/flux-sched` image used by CI has an older version of flux-core, and therefore the required `flux_shell_jobspec_info_unpack(3)` symbol is missing.

Maybe I can force a rebuild of the flux-sched CI so the `fluxrm/flux-sched` image will have the latest version of flux-core. If that works I'll rerun the CI here as well.

grondo:
> Maybe I can force a rebuild of the flux-sched CI

Well Github won't let me rerun the CI (maybe it has been too long or CI builds from pushes can't be re-run). There's a PR pending so once that is merged we can retry here.

SteVwonder (commented):
Thanks @jameshcorbett for putting this together! Awesome to have MPI launch working on the new systems.

jameshcorbett:
I get these two warnings, `libtool: warning: remember to run 'libtool --finish /g/g12/corbett8/local/rznevada/lib/flux/job-manager/plugins/` and `libtool --finish /g/g12/corbett8/local/rznevada/lib/flux/shell/plugins/`, I wonder can I make those part of the `make install` command?

t/t1001-cray-pals.t
```diff
@@ -0,0 +1,73 @@
...
+test_expect_success 'job-manager: pals port distributor works' '
+	jobid=$(flux mini submit -N2 -n2 true) &&
+	flux job wait-event -vt 15 ${jobid} cray_port_distribution &&
+	flux job wait-event -mports=[11999,11998] ${jobid} cray_port_distribution &&
+	flux job wait-event -vt 5 ${jobid} clean
+'
+
+test_expect_success 'job-manager: pals port distributor saves ports for multi-shell jobs' '
+	jobid=$(flux mini submit -n1 true) &&
+	! flux job wait-event -vt 15 ${jobid} cray_port_distribution
```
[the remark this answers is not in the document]
jameshcorbett: I think I have to leave the two in the last test because they're inside a Flux job? But I made the other changes.

jameshcorbett:
Thanks @SteVwonder for the review, I addressed your comments and cleaned up the commits quite a bit.

SteVwonder:
> I get these two warnings

Huh. I've never seen that error before.  Are you attempting to install to a different prefix that the configured prefix (https://stackoverflow.com/questions/32766609/libtool-installation-issue-with-make-install)?

SteVwonder (approved):
Thanks @jameshcorbett! LGTM!

jameshcorbett:
> > I get these two warnings
> 
> Huh. I've never seen that error before. Are you attempting to install to a different prefix that the configured prefix (https://stackoverflow.com/questions/32766609/libtool-installation-issue-with-make-install)?

No, I get it after a regular `make clean; autogen.sh; ./configure --prefix=blah; make; make install`.

grondo:
> No, I get it after a regular `make clean; autogen.sh; ./configure --prefix=blah; make; make install`.

I don't think you need to worry about `libtool --mode=finish` when installing plugins. Libtool is probably just complaining because the path you're installing the plugins into isn't in the standard or ld.so cache paths. You can safely ignore this warning.

grondo (approved):
Really impressive work @jameshcorbett!

Just a few pretty minor comments inline. I've approved assuming the trivial memory leak will get fixed, the rest of the comments are just suggestions.

:+1:

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -0,0 +1,262 @@
...
+
+    if (!(h = flux_future_get_flux (fut))) {
+        goto cleanup;
+    }
+    if (flux_kvs_lookup_get_unpack (fut,
+                                    "{s:{s:o}}",
+                                    "execution",
+                                    "nodelist",
+                                    &nodelist)
+            < 0
```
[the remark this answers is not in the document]
jameshcorbett: I don't like this either, but I was actually using flux-core's clang-format. https://github.com/flux-framework/flux-core/issues/3710

jameshcorbett:
If everything looks OK, could someone with write access set MWP?

grondo:
I added admin access to the @flux-framework/core team, of which I believe you are a member, so try to set the MWP label now?

jameshcorbett:
There doesn't appear to be a MWP label so I will just merge.

grondo:
Ah, you're right, the flux-coral2 repo does not yet appear to be set up with mergify.io.

Typically in flux-framework projects we don't merge our own PRs.

See [C4.1: Development Process](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#development-process) :

> Maintainers SHOULD NOT merge their own patches except in exceptional cases, such as non-responsiveness from other Maintainers for an extended period (more than 1-2 days).

Not a big deal in this case and I'm not even sure flux-coral2 is following C4.1, but just FYI...

SteVwonder:
> Ah, you're right, the flux-coral2 repo does not yet appear to be set up with mergify.io.

Good call.  Just opened #8

src/job-manager/plugins/cray_pals_port_distributor.c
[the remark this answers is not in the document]
trws: I'm looking over the clang-format file again, because I figured out how to get it to leave the custom formatting on pack/unpack alone.  This is configurable, but it's normally only done if the rest of the statement doesn't fit on the line.  It makes it so that long complex math expressions have their operator visible on the left rather than buried, but we can play with the costs and other things to see if we can get something better.  Do you have an example of how this should be formatted?

PR #10, opened 2021-09-30
DWS: Add `Workflow` creation & `NNF` watching

SteVwonder:
Add support for creating `Workflow` objects within K8s when a job is submitted with a `dw` jobspec attribute.  This functionality is tied to the existing `dws.create` RPC created by the jobtap plugin.

Add support for watching changes to `NearNodeFlash` (`NNF` for short) objects.  Currently, the script just logs to Flux's dmesg, but this functionality can be expanded in later PRs to mark NNFs as down within Fluxion or update their `percentDegraded` property within Fluxion.

Note: This PR builds upon #5, so that PR should go in first.  The first unique commit of this PR is 3fbfdf0.


src/modules/dws.py
```diff
@@ -94,6 +152,7 @@ def main():
     fh.reactor_run()
 
     w.stop().destroy()
+    del watchers
```
jameshcorbett, line 155: For this I think it would be better to use a context manager, since it's a little clearer and more robust (I'm actually not sure that this guarantees that `__del__` will be called even though it's the only reference). I think you could just rename `Watchers.__del__` to `Watchers.__exit__` and return `False` and then also add a simple `__enter__` method.

jameshcorbett (commented):
LGTM I only had one comment. I still want to play around with it in the VM though and make sure everything checks out, then I'll approve.

src/modules/dws.py
SteVwonder: Thanks! Just forced pushed a fixup commit with that change.

dongahn:
Great! Merging.

PR #14, opened 2021-11-05
build: drop ax_python_devel.m4

garlick:
Problem: ax_python_devel.m4 is not used.

Since flux-coral2 doesn't use PYTHON_CFLAGS or PYTHON_LIBS,
it would seem that there is no need for ax_python_devel.m4.
Furthermore, it has a bug that causes it to reject python
3.10.0 as too old.

Just drop ax_python_devel.m4 and adjust configure.ac
accordingly.


grondo (approved):
LGTM, Thanks! But possibly you want @jameshcorbett to give a final review.

jameshcorbett (approved):
LGTM as well. We don't have mergify.io set up so I will go ahead and merge if that's OK with you @garlick

garlick:
Yep

PR #18, opened 2022-02-01
build/test: rename centos8 to el8, set automake foreign mode and drop redundant files

grondo:
Now that flux-sched build has been updated to rename `centos8` to `el8`, do the same here.

Additionally add the `foreign` option to automake as has been done in other flux-framework projects and drop unnecessary files.

This touches the mergify config and thus will have to be manually merged.


jameshcorbett (approved):
LGTM. OK to merge?

grondo:
Yes, thanks!

Issue #16, opened 2022-01-14
Interference with flux-core and flux-sched testsuites


jameshcorbett:
Headaches can result from running flux-core and flux-sched testsuites when flux-coral2 is installed. When it is installed, the ``cray_pals`` shell plugin is always loaded, but the necessary ``cray_pals_port_distributor`` jobtap plugin is not. Without the jobtap plugin, all jobs will fail, and the error [is not always apparent](https://github.com/flux-framework/flux-sched/issues/899). It would be nice if there were a way to enforce that the shell plugin and the jobtap plugin go together.

Besides, if you specify that the jobtap plugin should be loaded with a configuration file, some flux-sched and flux-core tests will still fail.

grondo:
This could be a testsuite bug. The testsuite should sanitize the environment such that external plugins, modules, rc scripts, etc are not loaded. I'm assuming the `cray_pals` shell plugin is loaded due to `FLUX_SHELL_RC_PATH`? In any event, this environment variable should be removed in the testuites.

grondo:
Oops, I wasn't thinking clearly. `FLUX_SHELL_RC_PATH` is already unset in the flux-core testsuite:
https://github.com/flux-framework/flux-core/blob/76f00c5855b662bfbaf6d05253cf379ea22d564f/t/sharness.d/flux-sharness.sh#L378-L381

If flux-core tests are failing due the `cray_pals` shell plugin being loaded, I'd like to understand how, since that means some kind of external/installed environment is creeping into the test environment.

For flux-sched, the tests use the installed flux-core, so I could see here how `cray_pals` is being loaded automatically. However, the installed flux-core should also load the requisite jobtap plugin, otherwise the installed version won't work (as you describe above). Therefore, we should run down why installed flux isn't working.

If the `cray_pals` plugin only needs to be loaded in the system instance, we may need to figure out a way to only load the plugins for that instance, and not every batch job. What is the strategy for loading the plugins at this point?

jameshcorbett:
Turns out the flux-core failure was just due to ``FLUX_CONF_DIR`` being set in my environment and should be fixed by flux-core#4059. 

The cray-pals jobtap/shell plugins need to be loaded for all levels of the hierarchy, not just the system instance (by contrast the DWS plugins only need to be loaded by the system instance).

[...]

PR #19, opened 2022-02-07
cray-libpals: allow shell plugin to detect missing port-distributor jobtap plugin

grondo:
This PR adds better synchronization between the cray-pals shell plugin and the port-distributor jobtap plugin as discussed in #16. The jobtap plugin now emits the `cray_port_distribution` event under a `prolog` action, allowing the shell plugin to definitively detect a missing jobtap plugin if this event does not appear before the `start` event.

This means the shell plugin can be loaded without the jobtap plugin and jobs will still run successfully (e.g. if running the flux-sched testsuite)

One minor issue: This change doesn't prevent the shell plugin from setting most of the libpals environment variables (besides `PMI_CONTROL_PORT`) and writing out the apinfo file, since these actions happen before the plugin fetches the `cray_port_distribution` event. I figure this isn't a critical problem for now, since running with this package installed but without the jobtap plugin loaded will likely be an uncommon case.

### DIFF
--- a/src/job-manager/plugins/cray_pals_port_distributor.c
+++ b/src/job-manager/plugins/cray_pals_port_distributor.c
@@ -131,6 +131,12 @@ static void count_job_shells (flux_future_t *fut, void *arg)
     }
 
 cleanup:
+    if (flux_jobtap_prolog_finish (triple->plugin,
+                                   triple->jobid,
+                                   "cray-pals-port-distributor",
+                                   0) < 0)
+        flux_log_error (h,
+                        "cray_pals_port_distributor: prolog_finish");
     if (hlist)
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
LGTM, thanks! 

As for the plugin still setting all the other PALS environment variables I think that is probably OK, even long-term. If the job uses Cray MPI it will fail due to the port not being specified; if the job doesn't use Cray MPI it doesn't matter whether the PALS environment variables are set properly or not. The only difference I foresee it making is what the error message you get is when the ports aren't specified. That could turn out to be important but it's not worth worrying too much about now.

grondo:
Thanks! I'll set MWP

PR #23, opened 2022-08-18
DWS: update JGF writer for fluxion changes

jameshcorbett:
Problem: Fluxion added support for properties, and
added a new required argument to some Python functions.

Solution: add empty arguments to those functions.
Also update test baseline for JGF writer.


grondo (approved):
LGTM! I wonder if we should add a test run of flux-coral2 to the flux-sched CI (like we have for flux-accounting and flux-sched in the flux-core CI)

jameshcorbett:
That would be nice, I will make an issue in flux-sched.

Issue #27, opened 2022-08-30
libpals doesn't play well with attempting to oversubscribe cores


ryanday36:
A user was asking about running multiple MPI tasks per core on Tioga. It should be possible per  https://flux-framework.readthedocs.io/en/latest/faqs.html#how-can-i-oversubscribe-tasks-to-resources-in-flux. It works on corona:
```
[day36@corona178:~]$ flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free      1       48        8 corona178
 allocated      0        0        0 
      down      0        0        0
[day36@corona178:mpi_hello]$ flux mini run -n 48 -o per-resource.type=core -o per-resource.count=2 ./hello_corona
Number of MPI tasks is: 96

[day36@corona178:mpi_hello]$
```
but on tioga:
```
[day36@tioga24:mpi_hello]$ flux mini run -n64 -o per-resource.type=core -o per-resource.count=2 ./hello_tioga 
0.044s: flux-shell[0]: ERROR: libpals: taskid 64 (on node 0) >= ntasks 64
0.044s: flux-shell[0]: ERROR: libpals: Error initializing pals_pe_t structs
0.044s: flux-shell[0]: ERROR: plugin 'libpals': shell.init failed
0.044s: flux-shell[0]: FATAL: shell_init
0.044s: job.exception type=exec severity=0 shell_init
flux-job: task(s) exited with exit code 1
[day36@tioga24:mpi_hello]$
```

ryanday36:
@jameshcorbett: did this make it into a release or Buildfarm RPM? I'm still seeing the same error on Tioga, so I'm trying to figure out if it's still an issue or if Tioga just didn't get the update. It looks like Tioga has flux-coral2-0.1.0-1.t4.x86_64.

jameshcorbett:
Not into a release yet. Since you're asking though I will try to get a release out this week! I am a bit unclear on how updates work---should I try to get releases for this repo out before the TOSS updates or should I just make releases whenever it makes sense?

ryanday36:
Great! Thanks James. Ben Casses and the folks doing acceptance testing are looking to be able to overcommit nodes on Tioga. In general, getting releases out before TOSS updates is good. I think it's easier for them to do out of cycle updates on this repo though, as it's not in the main TOSS image.

PR #30, opened 2022-09-07
Pals incorrect task counts

jameshcorbett:
Problem: the libpals shell plugin does not account for task counts
not specified in the usual task-count jobspec position. For instance,
it does not count tasks specified with the per-resource.* shell options.

Solution: properly count tasks by checking the counts given by the shell
rather than the counts given in the jobspec.

Fixes https://github.com/flux-framework/flux-coral2/issues/27.

### DIFF
--- a/src/shell/plugins/cray_pals.c
+++ b/src/shell/plugins/cray_pals.c
@@ -554,33 +554,14 @@ cleanup:
     return ret;
 }
 
-static int get_ntasks (flux_shell_t *shell)
+static int get_cores_per_task (flux_shell_t *shell, int ntasks)
 {
-    int version, ntasks;
+    int version, task_slots, cores_per_slot;
 
     if (flux_shell_jobspec_info_unpack (shell,
-                                        "{s:i, s:i}",
+                                        "{s:i, s:i, s:i}",
                                         "version",
                                         &version,
-                                        "ntasks",
-                                        &ntasks) < 0
-        || version != 1) {
-        shell_log_error ("Error unpacking 'ntasks' from jobspec");
-        return -1;
-    }
-    return ntasks;
-}
-
-static int get_cores_per_task (flux_shell_t *shell)
-{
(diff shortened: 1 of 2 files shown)


garlick (commented):
Hey, before I forget I wanted to mention that here may be an unintentional copy of this branch pushed to the main repo:

https://github.com/flux-framework/flux-coral2/tree/pals-incorrect-task-counts

jameshcorbett:
Ah thanks. Deleted.

garlick (approved):
This seems right although I admit I had lost my bearing between shell ranks, task ranks, etc.  To convince myself it was alright I did install this on a test system and poke at it.  All good.

I guess we have no way to check the contents of the apinfo file.  That is kind of too bad - for extra credit sometime, it may be good to write a little python test program that pulls the data back out of that file and and emits it as json on stdout so it can then be checked by a sharness test using jq.

But that need not be done here.

jameshcorbett:
Thanks, set MWP. I will make a pass at a C test program---will have to be C and not Python because the contents of the file is a bunch of structs written out with no formatting or encoding.

grondo (approved):
This does work I think. It is indeed confusing that `ntasks` from `flux_shell_info_unpack(3)` is the total number of tasks in the job, while `flux_shell_rank_info_unpack(3)` returns `ntasks` as the number of local tasks. Sorry about that.

Note that due to the addition of `--tasks-per-core=N` to `flux mini` you could have more tasks than cores, not sure if that case is handled here.

garlick:
Oh the test proggie, up to you but it might be possible to in python:
  https://stackoverflow.com/questions/8710456/reading-a-binary-file-with-python

One of the answers references this
  https://docs.python.org/3/library/struct.html#struct.unpack

grondo:
Ok, that sounds good then! Just wanted to make sure `0` was acceptable and you didn't need `min(1, cores_per_task)`

PR #32, opened 2022-10-10
build: change shell plugin install directory

jameshcorbett:
Problem: flux-coral2 is being added to mainline TOSS. The cray-pals plugin is built into the main shell pluginpath, so it is loaded by default. This introduces some performance overhead.

Solution: move the shell plugin install directory into a subdirectory of the shell pluginpath so it is not
loaded by default.


grondo (approved):
LGTM!

PR #37, opened 2023-01-31
build: add DWS python files to build

jameshcorbett:
Problem: the build system doesn't know anything about the Python scripts and packages used for interacting with DWS. This means they are not included when running 'make dist' and are not installed.

Add the Python modules to the build.


grondo (approved):
LGTM!

PR #38, opened 2023-02-01
build: add missing header file

jameshcorbett:
Problem: a header file is missing from the build system, and it is not included in distributions.

Add the missing file.


jameshcorbett:
Not a priority @grondo , I'm just going to patch the RPM instead of making another release.

grondo (approved):
LGTM. BTW, adding `make distcheck` to CI might help catch this sort of thing in the future!

PR #40, opened 2023-02-03
dws: add jobid to RPC error message

jameshcorbett:
Problem: some error messages are logged which do not include the jobid, making it difficult or impossible to figure out which job is associated with the error.


cmoussa1 (approved):
LGTM!

PR #46, opened 2023-03-21
Dws jobtap epilog

jameshcorbett:
Flux jobs with rabbit file systems should be held in jobtap epilog while data movement is completed. Add an epilog to the dws-jobtap plugin and associated tests.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -26,6 +26,7 @@
 
 #define CREATE_DEP_NAME "dws-create"
 #define SETUP_PROLOG_NAME "dws-setup"
+#define DWS_EPILOG_NAME "dws-epilog"
 
 struct create_arg_t {
     flux_plugin_t *p;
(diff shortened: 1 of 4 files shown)


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -380,6 +462,16 @@ static int exception_cb (flux_plugin_t *p,
         }
         *prolog_active = 0;
     }
+    if ((epilog_active = flux_jobtap_job_aux_get (p, FLUX_JOBTAP_CURRENT_JOB, "dws_epilog_active")) && (*epilog_active)){
+        if (flux_jobtap_epilog_finish (p, id, SETUP_PROLOG_NAME, -1) < 0) {
```
grondo, line 466: Should this be `DWS_EPILOG_NAME`? Tests are working though, so I must not be understanding something here or this code path is not tested...

grondo (approved):
This looks pretty good to me! One question inline.

Also, when looking at the code it occurred to me it might be easier to modify `create_args` to contain `prolog_active` and `epilog_active` members, so you do not have to individually malloc integers and stash them in the job. (This would require that `create_args` persist for the duration of the job though, so maybe that isn't a net win. I leave it up to you.

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: You understood, it's just not tested. The code path is only covered when an exception is raised while the prolog is active. I'll have to add test cases for that---that's important.

I'll force-push this with the fix and merge, then make another PR for the tests.

PR #48, opened 2023-03-22
Add 'make distcheck' to CI

jameshcorbett:


garlick (approved):
Looks good!   If you get impatient later you could tweak this so that distcheck only runs on one of the builders instead of both, but looks fine to me for now.

PR #49, opened 2023-03-22
dws: kill workflows in error for too long

jameshcorbett:
Problem: workflows in Error state are immediately killed. However, that state is not always fatal.

Instead of immediately killing all jobs associated with Error workflows, only kill them after a configurable number of seconds have elapsed.


cmoussa1 (approved):
LGTM! Thanks @jameshcorbett.

PR #50, opened 2023-03-23
test: replace 'flux mini' usage

jameshcorbett:
Problem: 'flux mini CMD' is deprecated in favor of 'flux CMD'.

Replace usage in the test suite.


garlick (approved):
LGTM!   Hopefully this won't hinder testing on the actual systems before flux-core-0.48 is rolled out?  (Could always leave this PR pending until systems have caught up I guess)

cmoussa1 (approved):
LGTM too! I just realized I should probably do this in flux-accounting's test suite too... thanks for the inspiration!

PR #51, opened 2023-03-25
Dws prolog exceptions

jameshcorbett:
dws-jobtap.c adds jobtap prologs and epilogs to hold jobs in a state while data is moved to rabbits (the prolog) and from them (the epilog). If a severity-0 exception occurs during the prolog, the prolog should be removed immediately rather than waiting for the data movement to complete. However, there are no tests for that functionality. Also, epilogs are currently removed from jobs when exceptions occur, but there is no reason for that to be done, since the epilog runs once the job has already finished.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -138,7 +138,7 @@ static int depend_cb (flux_plugin_t *p,
     json_t *resources;
     json_t *jobspec;
     int userid;
-    int *prolog_active;
+    int *prolog_active = NULL;
 
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
@@ -165,6 +165,7 @@ static int depend_cb (flux_plugin_t *p,
                                      "dws_prolog_active",
                                      prolog_active,
                                      free) < 0){
+            free (prolog_active);
             flux_log_error (h, "dws-jobtap: error initializing exception-monitoring");
             return -1;
         }
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -453,7 +453,7 @@ static int exception_cb (flux_plugin_t *p,
         return -1;
     }
     if ((prolog_active = flux_jobtap_job_aux_get (p, FLUX_JOBTAP_CURRENT_JOB, "dws_prolog_active")) && (*prolog_active)){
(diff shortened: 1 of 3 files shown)


jameshcorbett:
@grondo CI is passing now, could I get you to take a look?

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -463,7 +463,7 @@ static int exception_cb (flux_plugin_t *p,
         *prolog_active = 0;
     }
     if ((epilog_active = flux_jobtap_job_aux_get (p, FLUX_JOBTAP_CURRENT_JOB, "dws_epilog_active")) && (*epilog_active)){
-        if (flux_jobtap_epilog_finish (p, id, DWS_EPILOG_NAME, -1) < 0) {
+        if (flux_jobtap_epilog_finish (p, id, DWS_EPILOG_NAME, 1) < 0) {
             flux_log_error (h,
                             "Failed to finish epilog %s for job %" PRIu64 " after exception",
                             DWS_EPILOG_NAME,
```
garlick: Is this change part of adding the test?  I guess it's still reporting failure, just a code of +1 instead of -1?  Why?

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -163,21 +160,15 @@ static int depend_cb (flux_plugin_t *p,
...
                                      prolog_active,
-                                     free) < 0
-            || flux_jobtap_job_aux_set (p,
-                                     FLUX_JOBTAP_CURRENT_JOB,
-                                     "dws_epilog_active",
-                                     epilog_active,
                                      free) < 0){
             flux_log_error (h, "dws-jobtap: error initializing exception-monitoring");
             return -1;
         }
```
garlick, line 170: Looks like if malloc succeeds but aux_set fails, `prolog_active` is leaked?

garlick (approved):
This looks fine to me, although I'm not that familiar with the prolog/epilog stuff.
Just a couple minor comments.  I'll tentatively approve.

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: Yeah it's part of the test, which checks that the prolog exits with a failure and an exit code of 1. I think `-1` would be fine as well but there are other parts of the plugin that finish the prolog/epilog with `1` rather than `-1`. It's solely for consistency.

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: Good catch, will fix.

PR #53, opened 2023-04-04
Rabbit selection based on compute nodes

jameshcorbett:
Due to some issues with Fluxion, rabbits can't be scheduled properly just yet. A workaround is needed for the systems that have rabbits currently (Tioga and Hetchy, and soon Tenaya and RZVernal). The way it works on hetchy currently is a shameless hack.

A better solution for now is to keep the scheduler ignorant of the rabbits, and make scheduling decisions based on normal factors like cores and nodes. Then, pick rabbits based on the set of compute nodes chosen.


cmoussa1 (approved):
This LGTM! Thanks @jameshcorbett

PR #54, opened 2023-04-04
Split DW directives across lines

jameshcorbett:
The first two commits are minor tweaks to improve error messages.

Requests for rabbit storage are strings that look like `#DW jobdw capacity=10GiB type=xfs name=project1`. The strings go either as a string or a list of strings into jobspec `attributes.system.dw`. Users might therefore do something like `flux submit --setattr=system.dw="#DW jobdw capacity=10KiB type=xfs name=project1" -n1 sleep 15`. If they want multiple directives (which will be common) they currently need to do something like `flux submit --setattr=^system.dw="xfs_copy_in.json"` where `xfs_copy_in.json` is a JSON file containing a list of DW directives. 

By splitting DW directives across lines, we can allow users to write things like `flux submit --setattr=system.dw="#DW jobdw capacity=10KiB type=xfs name=project1 \n #DW jobdw capacity=10KiB type=xfs name=project1" -n1 sleep 15`, thereby stuffing two directives into the command line. Doesn't look like a huge improvement, but it makes the following legal in a batch script, which is more important:

```
#!/bin/bash

#flux: -N1
#flux: --setattr=system.dw="""
#flux: #DW jobdw capacity=10GiB type=xfs name=project1
#flux: #DW copy_in source=/p/lslide/corbett8/ destination=$DW_JOB_project1/
#flux: """
```

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -101,7 +101,9 @@ def create_cb(fh, t, msg, api_instance):
     jobid = msg.payload["jobid"]
     userid = msg.payload["userid"]
     if isinstance(dw_directives, str):
-        dw_directives = dw_directives.splitlines()
+        # assume different directives are on different lines and remove
+        # any blank lines
+        dw_directives = [dw for dw in dw_directives.splitlines() if dw]
     if not isinstance(dw_directives, list):
         raise TypeError(
             f"Malformed dw_directives, not list or string: {dw_directives!r}"
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -101,7 +101,7 @@ def create_cb(fh, t, msg, api_instance):
     jobid = msg.payload["jobid"]
     userid = msg.payload["userid"]
     if isinstance(dw_directives, str):
-        dw_directives = [dw_directives]
+        dw_directives = dw_directives.splitlines()
     if not isinstance(dw_directives, list):
         raise TypeError(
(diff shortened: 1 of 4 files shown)


cmoussa1 (approved):
This LGTM too! Nice work

jameshcorbett:
@cmoussa1 could I get you to re-review quickly? I made some minor changes and tacked on an extra commit.

cmoussa1:
Sure thing! I'll give it a look now. Sorry for the late response, I was getting my laptop screen replaced this morning.

src/modules/coral2_dws.py
```diff
@@ -101,7 +101,9 @@ def create_cb(fh, t, msg, api_instance):
     jobid = msg.payload["jobid"]
     userid = msg.payload["userid"]
     if isinstance(dw_directives, str):
-        dw_directives = dw_directives.splitlines()
+        # assume different directives are on different lines and remove
+        # any blank lines
+        dw_directives = [dw for dw in dw_directives.splitlines() if dw]
```
cmoussa1, line 106: IMHO this commit could be squashed into your third commit, `dws: split DW directives by newline` since it looks like it also makes a change to this line of code. I'll leave that decision up to you though. 🙂

cmoussa1 (approved):
This LGTM! Just one minor comment about potentially squashing a commit. Nice job!

src/modules/coral2_dws.py
jameshcorbett: Good point, thanks. I'll squash and set MWP.

PR #56, opened 2023-04-04
Fix workflow error handling

jameshcorbett:
Workflows that are stuck in "State: Error" should be killed off after a configurable number of seconds. But the logic for doing so is broken. Fix the logic and add a test.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -53,6 +53,7 @@ class WorkflowInfo:
         self.computes = None
         self.breakdowns = None
         self.last_error_time = None
+        self.last_error_message = None
 
 
 def message_callback_wrapper(func):
@@ -330,9 +331,11 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, fh, k8s_api):
         )
         if winfo.last_error_time is None:
             winfo.last_error_time = time.time()
+        winfo.last_error_message = workflow["status"].get("message", "")
         WORKFLOWS_IN_ERROR.add(winfo)
     else:
         winfo.last_error_time = None
+        winfo.last_error_message = None
         WORKFLOWS_IN_ERROR.discard(winfo)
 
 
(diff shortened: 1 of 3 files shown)


src/modules/coral2_dws.py
```diff
@@ -400,7 +403,8 @@ def kill_workflows_in_error(reactor, watcher, _r, error_timeout):
                 winfo.jobid,
                 "exception",
                 0,
-                "DWS/Rabbit interactions failed: workflow stuck in Error too long",
+                "DWS/Rabbit interactions failed: workflow in 'Error' state too long: "
+                f"{winfo.last_error_message}",
```
cmoussa1, line 407: minor nit on the commit message: i think `workflows` should be `workflow` in the commit message.

t/t1002-dws-workflow-obj.t
```diff
@@ -126,4 +126,14 @@ test_expect_success 'job-manager: dependency plugin works when validation fails'
...
+test_expect_success 'workflows in Error are killed properly' '
+	jobid=$(flux submit --setattr=system.dw="#DW jobdw capacity=10KiB type=xfs name=project1
+			#DW copy_in source=/some/fake/dir destination=$DW_JOB_project1/" \
+		    -N1 -n1 hostname) &&
+	flux job wait-event -vt 10 -m description=${CREATE_DEP_NAME} \
+		${jobid} dependency-add &&
+	flux job wait-event -t 10 ${jobid} exception &&
+	flux job wait-event -vt 5 ${jobid} clean
+'
+
```
cmoussa1, line 138: I think there might be a whitespace inconsistency here. Feel free to just ignore this if I am wrong or missing something; just looks like the second and third lines from the first command in this test are indented two and one tab(s) too many.

cmoussa1 (approved):
This LGTM! Couple of tiny comments about a commit message and some whitespace stuff.

You might also consider adding a short PR description now that #54 is merged and this is no longer in [WIP].

t/t1002-dws-workflow-obj.t
jameshcorbett: Good catch, thanks. There are some tabs and spaces mixed together, and that makes it look bad when the tab size changes. I'll make it consistently tabs.

PR #58, opened 2023-04-04
load port distributor in all Flux instances, and support -o pmi=cray-pals

### LINKED ISSUE #52: jobtap plugin not loaded: disabling operation
I happened to notice on rzvernal that this message appears when launching slurm -> flux -> cray MPI program:
```
rzvernal14 /g/g0/garlick > srun -N2 -n2 --mpi=pmi2 flux start flux run -o mpi=none -o verbose=2 -N2 proj/mpi-hello/hello
[snip]
0.062s: flux-shell[1]: DEBUG: libpals: jobtap plugin not loaded: disabling operation
[snip]
```
It's certainly expected behavior that the jobtap plugin is only loaded in the system instance, while the shell plugin gets loaded in all instances (at least when flux is run from the system path).

What's happening here?   How does the `cray_pals.so` shell plugin function when it cannot reach the `cray_pals_port_distributor.so` jobtap plugin?

My MPI program seems to be working despite the jobtap plugin not being loaded.

Note that due to flux-framework/flux-core#5045 and recent environment changes on LC coral2 systems, Cray MPI programs will no longer find the flux PMI2 library before the Cray one.  I just want to make sure that everything is going to keep working!

garlick:
This resolves #52 by adding an rc script that loads the port distributor jobtap plugin in all instances.

In addition it makes a slight tweak to the cray_pals plugin so that it conforms to the way the `pmi` shell option is being changed in flux-core, as described in flux-framework/flux-core#5063.

These changes assume that `flux-coral2` is built with the same prefix as `flux-core`.

Also, we'll want to ask the sys admins to
1) drop the job manager config in the system instance that loads the port distributor plugin
2) drop the change to the shell initrc.lua that loads the cray_pals plugin from its side-installed location
3) add a change to the shell initrc.lua to add `cray-pals` to the default set of PMIs offered to jobs

The last one looks like this
```lua
-- Add cray-pals to the default PMIs offered on this system
if shell.options['pmi'] == nil then
    shell.options['pmi'] = 'cray-pals,simple'
end
```
(It needs to be before the line that loads all hte `*.so` plugins)

I've been testing this on my home system and it all seems to go together well.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,3 +44,7 @@ check-prep: all
 	cd src && $(MAKE) check
 	cd doc && $(MAKE) check
 	cd t && $(MAKE) check-prep
+
+
+dist_fluxrc1_SCRIPTS = \
+	etc/01-coral2-rc
--- a/configure.ac
+++ b/configure.ac
@@ -113,12 +113,6 @@ AC_SUBST(fluxrc1dir)
 AS_VAR_SET(fluxrc3dir, $sysconfdir/flux/rc3.d)
 AC_SUBST(fluxrc3dir)
 
-AS_VAR_SET(fluxrc1dir, $sysconfdir/flux/rc1.d)
-AC_SUBST(fluxrc1dir)
-
-AS_VAR_SET(fluxrc3dir, $sysconfdir/flux/rc3.d)
-AC_SUBST(fluxrc3dir)
-
 AS_VAR_SET(fluxk8spydir, $pyexecdir/flux_k8s)
 AC_SUBST(fluxk8spydir)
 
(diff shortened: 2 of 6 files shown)


jameshcorbett:
I think it's actually because new flux-sched images haven't been built. I'll try building locally.

jameshcorbett (approved):
Code changes look great, thanks! Now just need to figure out how to get CI to pass. I'm still building locally.

jameshcorbett:
@garlick CI is passing! I'll let you rebase and set MWP whenever you're ready?

garlick:
Great! Will do.

PR #60, opened 2023-04-10
dws: install shell plugin to main dir

jameshcorbett:
Problem: the dws_environment shell plugin is not
loaded by default. The plugin has no effect when
a job has no DW string, so there is no drawback
to always loading it.

Install the plugin to the shell plugin dir, not
a subdirectory.


grondo (approved):
Looks fine to me. Is there a test that ensures normal jobs run fine with the `dws_environment` plugin loaded?

jameshcorbett:
> Looks fine to me. Is there a test that ensures normal jobs run fine with the `dws_environment` plugin loaded?

Yeah, there is. 

Setting MWP.

PR #61, opened 2023-04-14
systemd: change Requires to BindTo

### LINKED ISSUE #59: systemd: use `BindsTo=flux.service` instead of `Requires=flux.service`
As noted in flux-framework/flux-core#5072 and flux-framework/flux-accounting#337, we found that `Requires=flux.service` in systemd unit files of flux "service" scripts is causing flux to restart unexpectedly when it should stay shut down.

jameshcorbett:
Problem: the coral2-dws systemd service 'requires' flux which means (in combination with 'Restart=always') that the flux service cannot be shut down without first shutting down the coral2-dws service. That is obnoxious.

Solution: change 'requires' to 'bindsto' which is
a looser requirement.

Fixes #59.

### DIFF
--- a/etc/flux-coral2-dws.service.in
+++ b/etc/flux-coral2-dws.service.in
@@ -1,6 +1,6 @@
 [Unit]
 Description=Flux-DWS communication service
-Requires=flux.service
+BindsTo=flux.service
 After=flux.service
 
 [Service]


grondo (approved):
Looks good!

PR #63, opened 2023-04-21
dws: split directives by DW token

jameshcorbett:
Problem: coral2_dws.py splits DW strings by
newlines. This generally works, but is restrictive.

Split DW strings by the '#DW ' token. This should
prove both more robust and more flexible.


grondo (approved):
Looks fine to me!

PR #68, opened 2023-05-01
Unset PALS_ variables by default in shell plugin

### LINKED ISSUE #67: Unset PALS_ environment variables by default?
I noticed that when trying to run a Cray MPI application with `opmi=none`, it crashed and I'm guessing it was because it picked up the `PALS_` variables set in my interactive allocation. I'm wondering if maybe the `cray_pals` shell plugin should always unset the `PALS_` variables, even if `cray-pals` isn't in the PMI options (i.e. even if the plugin is deactivated). Thoughts @garlick ?

```
[corbett8@hetchy16:Flux]$ flux run -N2 -n2 -opmi=none ./mpi_hello_hetchy 
Fri Apr 28 23:00:59 2023: [unset]:_pmi_pals_init:pals_init2() failed: 2
Fri Apr 28 23:00:59 2023: [unset]:handle_launcher_type:_pmi_pals_init returned -1
MPICH ERROR [Rank 0] [job id unknown] [Fri Apr 28 23:00:59 2023] [hetchy17] - Abort(1091855) (rank 0 in comm 0): Fatal error in PMPI_Init: Other MPI error, error stack:
MPIR_Init_thread(170): 
MPID_Init(441).......: 
MPIR_pmi_init(110)...: PMI_Init returned 1

aborting job:
Fatal error in PMPI_Init: Other MPI error, error stack:
MPIR_Init_thread(170): 
MPID_Init(441).......: 
MPIR_pmi_init(110)...: PMI_Init returned 1
^Cflux-job: one more ctrl-C within 2s to cancel or ctrl-Z to detach
^C21.977s: job.exception type=cancel severity=0 interrupted by ctrl-C
flux-job: task(s) exited with exit code 255
[corbett8@hetchy16:Flux]$ flux run -N2 -n2 -opmi=none env | grep PALS
PALS_NODEID=0
PALS_RANKID=0
PALS_APINFO=/var/tmp/corbett8/jobtmp-0-fcDF3n9WfqZ/libpals_apinfo
PALS_NODEID=0
PALS_SPOOL_DIR=/var/tmp/corbett8/jobtmp-0-fcDF3n9WfqZ
PALS_APID=261535830416294912
PALS_RANKID=0
PALS_APINFO=/var/tmp/corbett8/jobtmp-0-fcDF3n9WfqZ/libpals_apinfo
PALS_SPOOL_DIR=/var/tmp/corbett8/jobtmp-0-fcDF3n9WfqZ
PALS_APID=261535830416294912
[corbett8@hetchy16:Flux]$ echo $PALS_RANKID
0
```

jameshcorbett:
Fixes #67 .

### DIFF
--- a/src/shell/plugins/cray_pals.c
+++ b/src/shell/plugins/cray_pals.c
@@ -758,6 +758,21 @@ static bool member_of_csv (const char *list, const char *name)
     return false;
 }
 
+/*
+ * Unset all PALS_* variables.
+ */
+static int unset_pals_env (flux_shell_t *shell)
+{
+    char *pals_env[] = { "PALS_NODEID", "PALS_RANKID",
+                         "PALS_APINFO", "PALS_APID",
+                         "PALS_SPOOL_DIR", "PMI_CONTROL_PORT" };
+    for (int i = 0; i < sizeof (pals_env) / sizeof (pals_env[0]); i++)
+    {
+        flux_shell_unsetenv (shell, pals_env[i]);
+    }
+    return 0;
+}
+
 int flux_plugin_init (flux_plugin_t *p)
 {
     const char *pmi_opt = NULL;
@@ -772,7 +787,9 @@ int flux_plugin_init (flux_plugin_t *p)
         return -1;
     }
     if (!pmi_opt || !member_of_csv (pmi_opt, "cray-pals"))
-        return 0; // plugin disabled
+        // plugin disabled, unset all inherited PALS_ variables
+        // otherwise libPALS might pick up and try to use them
+        return unset_pals_env (shell);
 
     shell_debug ("enabled");
 
--- a/t/t1001-cray-pals.t
+++ b/t/t1001-cray-pals.t
@@ -12,6 +12,8 @@ test_under_flux 2 job
 
 flux setattr log-stderr-level 1
 
+unset PALS_RANKID PALS_NODEID PMI_CONTROL_PORT
+
 test_expect_success 'job-manager: load cray_pals_port_distributor plugin with invalid config' '
 	test_expect_code 1 flux jobtap load ${JOBTAP_PLUGINPATH}/cray_pals_port_distributor.so \
 		port-min=0 port-max=12000 &&
@@ -75,6 +77,15 @@ test_expect_success 'shell: cray-pals is active with -opmi includes cray-pals' '
 	    printenv PALS_RANKID &&
 	flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=cray-pals,simple \
 	    printenv PALS_RANKID
+'
+test_expect_success 'shell: cray-pals unsets PALS variables when inactive' '
+	(export PALS_RANKID=0 PMI_CONTROL_PORT=6 && PALS_NODEID=1 &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PALS_RANKID &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PALS_NODEID &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PMI_CONTROL_PORT)
 '
 
 test_expect_success 'shell: pals shell plugin sets environment' '


jameshcorbett:
Whoops, looks like some tests failed. Investigating...

src/shell/plugins/cray_pals.c
```diff
@@ -758,6 +758,23 @@ static bool member_of_csv (const char *list, const char *name)
...
 
+/*
+ * Unset all PALS_* variables.
+ */
+static int unset_pals_env (flux_shell_t *shell)
+{
+    char *pals_env[] = { "PALS_NODEID", "PALS_RANKID",
+                         "PALS_APINFO", "PALS_APID",
+                         "PALS_SPOOL_DIR", "PMI_CONTROL_PORT" };
+    for (int i = 0; i < sizeof (pals_env); i++)
```
garlick, line 769: I think you want `sizeof (pals_env) / sizeof (pals_env[0])` here.  Or use the ccan ARRAY_SIZE() macro (preferred IMHO)

jameshcorbett: Thanks! That would explain the nondeterministic errors I was seeing. Where do I get `ccan` from though? It looks like it's vendored into flux-core?

garlick: Yeah you'd have to pull it in.  Maybe it's easier to just fix as above.

jameshcorbett:
I'm confused by the errors in CI:

```
  expecting success: 
  	unset PALS_RANKID &&
  	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=off \
  	    printenv PALS_RANKID
  
  0.039s: flux-shell[0]: ERROR: 
  0.039s: flux-shell[0]: ERROR: /usr/src/t/trash-directory.t1001-cray-pals/cray_pals.lua:4: plugin.load: /usr/src/src/shell/plugins/.libs/cray_pals.so: Success
  0.039s: flux-shell[0]: ERROR: /etc/flux/shell/initrc.lua:11: source /usr/src/t/trash-directory.t1001-cray-pals/cray_pals.lua failed
  0.039s: flux-shell[0]: FATAL: loading rc file /etc/flux/shell/initrc.lua
  0.040s: job.exception type=exec severity=0 loading rc file /etc/flux/shell/initrc.lua
  flux-job: task(s) Terminated
  test_must_fail: died by signal: flux run -o userrc=/usr/src/t/trash-directory.t1001-cray-pals/cray_pals.lua -o pmi=off printenv PALS_RANKID
```

And it's passing locally...

grondo:
Hm, yeah, I can't explain this non-error error:
```
/usr/src/t/trash-directory.t1001-cray-pals/cray_pals.lua:4: plugin.load: /usr/src/src/shell/plugins/.libs/cray_pals.so: Success
```
Sorry.

We may have to add debugging to the test, or add the tmate action and ssh in to the build container to see why this is failing.

I can try that.

grondo:
I think I see what is going on: `unset_pals_env()` returns `< 0` if any of the `PALS_*` environment variables are not found:

```c
static int unset_pals_env (flux_shell_t *shell)
{
    char *pals_env[] = { "PALS_NODEID", "PALS_RANKID",
                         "PALS_APINFO", "PALS_APID",
                         "PALS_SPOOL_DIR", "PMI_CONTROL_PORT" };
    for (int i = 0; i < sizeof (pals_env) / sizeof (pals_env[0]); i++)
    {
        if (flux_shell_unsetenv (shell, pals_env[i]) < 0) {
            return -1;
        }
    }
    return 0;
}
```

Since the plugin returns the value of this function from `flux_plugin_init()` 
```c
    if (!pmi_opt || !member_of_csv (pmi_opt, "cray-pals"))
        // plugin disabled, unset all inherited PALS_ variables
        // otherwise libPALS might pick up and try to use them
        return unset_pals_env (shell);
```

Then the plugin initialization fails and the plugin fails to load. Unfortunately, `flux_plugin_load()` sets an empty error here so we get a blank error message. I'll change that to something like "plugin init function failed".

I think here the return code from `flux_shell_unsetenv()` should be ignored so that all the environment variables have a chance to be removed, and the function should always return success.

jameshcorbett:
> I think I see what is going on: `unset_pals_env()` returns `< 0` if any of the `PALS_*` environment variables are not found:

> I think here the return code from `flux_shell_unsetenv()` should be ignored so that all the environment variables have a chance to be removed, and the function should always return success.

Thanks @grondo for looking into that! I thought that would explain it but it didn't make sense given that things were working for me locally. But I think I know why now. It seems that if I do a `make install` the test suite is affected. I had done a `make install` on `master` and then was running tests on my `unset-pals-env` branch (this PR) and everything worked. But when I ran `make install` on my branch and then ran the test suite, I got a lot of failures, presumably because the broken cray-pals plugin is always loaded by the shell initrc of the flux I was testing against, and loading the plugin was failing every job. Running `make uninstall` gave me back the desired behavior.

Is that an unavoidable gotcha or is there something I can do to insulate the test suite from the coral2 plugins added to the Flux install it's testing against?

In any case that's a separate issue and I think this PR is ready for review.

grondo:
> Is that an unavoidable gotcha or is there something I can do to insulate the test suite from the coral2 plugins added to the Flux install it's testing against?

Good question. I hadn't before considered that a framework project testing against installed flux-core would be influenced by the configuration of the installed Flux. Perhaps we need to open a flux-core issue on this because I'm having trouble at the moment thinking this through.

I guess one thing you could try is to disable the system shell initrc so that no extra plugins are loaded. One way to do this would be to set `-o initrc=testrc.lua`, where the `testrc.lua` would be empty or maybe contain just
```lua
if shell.options.userrc then source (shell.options.userrc) endb
```
since I think the flux-coral2 testsuite uses `-o userrc=...`. You could also try replacing uses of `-o userrc=` with `-o initrc=`.

I'm not sure if the shell plugin is the only gotcha here though. Perhaps there should be a better way to run a test instance in a "clean" environment for these testsuites, which is why I suggest a flux-core issue be opened.

jameshcorbett:
@garlick can you give this a look over now that things are passing?

t/t1001-cray-pals.t
```diff
@@ -59,23 +59,36 @@ test_expect_success 'shell: create shell initrc for testing' "
...
 	    printenv PALS_RANKID
 '
+test_expect_success 'shell: cray-pals unsets PALS variables when inactive' '
+	export PALS_RANKID=0 PMI_CONTROL_PORT=6 && PALS_NODEID=1 &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PALS_RANKID &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PALS_NODEID &&
+	test_must_fail flux run -o userrc=$(pwd)/$USERRC_NAME -o pmi=none \
+		printenv PMI_CONTROL_PORT
```
garlick, line 90: It may be advisable to enclose this whole test in parenthesis to force it to run in a sub-shell.  That prevents environment vars from affecting other tests.

The `unset PALS_RANKID` could  be done once before the tests begin, instead of in each test, as long as the tests are all careful not to set that variable.

src/shell/plugins/cray_pals.c
```diff
@@ -758,6 +758,21 @@ static bool member_of_csv (const char *list, const char *name)
     return false;
 }
```
garlick, line 760: In the commit message, in "have been inherited by a parent job", should "by" be replaced with "from"?

garlick (approved):
Looks good and this could go in as far as I'm concerned..  I just had a couple of comments for you to consider if you have time.

src/shell/plugins/cray_pals.c
jameshcorbett: Good catch, thanks!

PR #71, opened 2023-05-17
DWS kubeconfig command-line option

jameshcorbett:
The coral2_dws script runs as a systemd service as the flux user, but needs access to kubernetes. Access to kubernetes is generally achieved with [kubeconfig files](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/). Brian Behlendorf told me that placing the kubeconfig in the flux user's home directory (where kubernetes looks by default) would be inconvenient and that allowing the service's kubeconfig to be configured with a command-line argument would be ideal.

### DIFF
--- a/t/t1002-dws-workflow-obj.t
+++ b/t/t1002-dws-workflow-obj.t
@@ -34,18 +34,18 @@ test_expect_success 'job-manager: load dws-jobtap plugin' '
 
 test_expect_success 'exec dws service-providing script' '
 	R=$(flux R encode -r 0) &&
-	jobid=$(flux submit \
+	DWS_JOBID=$(flux submit \
 	        --setattr=system.alloc-bypass.R="$R" \
 	        -o per-resource.type=node --output=dws.out --error=dws.err \
 	        python ${DWS_MODULE_PATH} -e1) &&
-	flux job wait-event -vt 15 -p guest.exec.eventlog ${jobid} shell.start
+	flux job wait-event -vt 15 -p guest.exec.eventlog ${DWS_JOBID} shell.start
 '
 
 # This test used to close the race condition between the python process starting
 # and the `dws` service being registered.  Once https://github.com/flux-framework/flux-core/issues/3821
 # is implemented/closed, this can be replaced with that solution.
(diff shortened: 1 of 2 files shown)


t/t1002-dws-workflow-obj.t
```diff
@@ -157,4 +157,28 @@ test_expect_success 'workflows in Error are killed properly' '
...
+	jobid=$(flux submit --setattr=system.dw="#DW jobdw capacity=10KiB type=xfs name=project1" \
+		-N1 -n1 hostname) &&
+	flux job wait-event -vt 10 -m description=${CREATE_DEP_NAME} \
+		${jobid} dependency-add &&
+	flux job wait-event -vt 5 -m description=${PROLOG_NAME} \
+		${jobid} prolog-start &&
+	flux job wait-event -vt 5 -m description=${EPILOG_NAME} \
+		${jobid} epilog-start &&
+	flux job wait-event -vt 5 ${jobid} clean &&
+	flux job cancel ${DWS_JOBID}
```
cmoussa1, line 181: Is the `flux job cancel ${DWS_JOBID}` statement at the end of this test referencing the jobid that was created in the previous test? Or is it supposed to be `flux job cancel ${jobid}` here?

cmoussa1 (approved):
LGTM! Just one minor question/clarification (sorry if it's a dumb question, I could totally be misinterpreting the tests you wrote 😅).

t/t1002-dws-workflow-obj.t
jameshcorbett: It is referencing the jobid that was created in the previous test. That jobid is for a job spoofing a persistent service. It's not strictly necessary to cancel it because when the test suite ends it will be killed anyway, but I figured it couldn't hurt...

jameshcorbett:
Thanks! Setting MWP.

t/t1002-dws-workflow-obj.t
cmoussa1: That's what I figured at first, thanks for the clarification!!

PR #80, opened 2023-07-13
build: add `make deb` target for test packaging

garlick:
This adds a `make deb` target that builds a test debian package.

Why, you ask, when this project is specifically for a cray system that uses RPMs?   Well, it provides another way to check things out.  In particular, I found it useful to install this on a non-cray test system to make sure the `cray-pals` PMI plugin behaved as expected with the other PMI plugins, even though no client of `cray-pals` could be launched on that system.  Maybe in the future it can be of help when debugging the port distributor or future VNI support.


jameshcorbett (approved):
I'm not very familiar with Debian but I don't see anything harmful here!

garlick:
Cool, thanks!  I'll set MWP.

PR #81, opened 2023-07-20
DWS service crashes

### LINKED ISSUE #64: Handling coral2_dws crashes
When the coral2_dws.py script crashes, all the workflow objects that exist in kubernetes are orphaned. The script keeps a cache of workflows it recognizes, and only workflows that it has submitted itself (previous iterations of the script don't count) go in the cache. This is bad. If the script crashes regularly (even if infrequently) kubernetes will start filling up with workflows and users may see their jobs stuck in various states. In the worst case, a job may be stuck in the jobtap epilog, which is extremely obnoxious to remove.

The reason for this design is partly that the dws-jobtap plugin sends RPCs to coral2_dws script, which holds on to the RPCs and responds once certain conditions have been met. If the script crashes, I think (?) there would be no way to respond to the RPCs.

A simple solution would be for the coral2_dws script to kill off any existing workflows and their associated jobs when it starts up. But that would cause users a lot of grief, if their jobs have been sitting in the queue for a long time or (worse) their job has been running for a while or (worst) their job has completed and is just waiting for data movement to complete.

A better solution would be for coral2_dws to become stateless (or closer to it) and process workflows it has submitted itself the same way it treats workflows it finds in kubernetes that it didn't submit. The RPC-response model could be replaced by an RPC-RPC model where the jobtap plugin and the coral2_dws script send RPCs to each other, and don't expect any response. This would require assuming that when the coral2_dws script starts up, any workflows that exist are associated with flux jobs, and that processing them normally wouldn't cause any problems...

The jobtap plugin should be made to re-send its RPCs once after ~10 seconds if they fail the first time because the coral2_dws sevice may be down.

jameshcorbett:
This PR is a substantial rewrite of `dws-jobtap.c` and `coral2_dws.py` to solve the problem of handling `coral2_dws.py` crashes. Before this PR is merged, the `coral2_dws.py` service assumes that its lifetime encompasses the lifetime of any job that it sees. It can't handle rabbit jobs that existed before it started. This is because it keeps an internal data structure of active rabbit jobs, and jobs can only be added to the data structure when they are being created. 

Furthermore, the `dws-jobtap.c` plugin sends a RPC to `coral2_dws.py` when a job is ready (by flux-core standards) to transition states, and `coral2_dws.py` responds to that RPC when HPE rabbit software gives the go-ahead. But HPE rabbit software can take a considerable amount of time to give the go-ahead. If the `coral2_dws.py` service crashes or restarts in the meanwhile, it won't be able to respond to the RPC it received, and the job will be stranded. In the worst case, the job is stranded in the dws epilog, and can't be canceled. That is intensely obnoxious.

Instead of making the `coral2_dws.py` service hold on to RPCs, make it respond more or less immediately, and send a new RPC to the jobtap plugin once the go-ahead is given by HPE rabbit software.

Fixes #64

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -92,7 +92,6 @@ static void create_cb (flux_future_t *f, void *arg)
     const char *errstr = NULL;
     flux_t *h = flux_future_get_flux(f);
     struct create_arg_t *args = flux_future_aux_get (f, "flux::create_args");
-    json_t *jobspec = arg, *resources = NULL;
 
     if (args == NULL) {
         flux_log_error (h, "create args missing in future aux");
@@ -100,25 +99,15 @@ static void create_cb (flux_future_t *f, void *arg)
     }
 
     if (flux_rpc_get_unpack (f,
-                             "{s:b, s?s, s?o}",
+                             "{s:b, s?s}",
                              "success", &success,
-                             "errstr", &errstr,
-                             "resources", &resources) < 0)
+                             "errstr", &errstr) < 0)
     {
         raise_job_exception (h, args->id, CREATE_DEP_NAME, "Failed to unpack dws.create RPC");
         goto done;
     }
 
-    if (success && resources) {
-        if (flux_jobtap_dependency_remove (args->p, args->id, CREATE_DEP_NAME) < 0) {
-            raise_job_exception (h, args->id, CREATE_DEP_NAME, "Failed to remove dependency for job");
-            goto done;
-        }
-        if (json_object_set (jobspec, "resources", resources) < 0){
-            raise_job_exception (h, args->id, CREATE_DEP_NAME, "Failed to update jobspec with DWS resources");
-            goto done;
-        }
-    } else {
+    if (!success) {
         if (errstr){
             raise_job_exception (h, args->id, CREATE_DEP_NAME, errstr);
         }
@@ -128,7 +117,6 @@ static void create_cb (flux_future_t *f, void *arg)
     }
 
 done:
-    json_decref (jobspec);
     flux_future_destroy (f);
 }
 
@@ -190,11 +178,10 @@ static int depend_cb (flux_plugin_t *p,
         create_args->p = p;
         create_args->id = id;
         if (flux_future_aux_set (create_fut, "flux::create_args", create_args, free) < 0
-            || flux_future_then (create_fut, -1, create_cb, jobspec) < 0) {
+            || flux_future_then (create_fut, -1, create_cb, NULL) < 0) {
             flux_future_destroy (create_fut);
             return -1;
         }
-        json_incref (jobspec); // create_fut will use and destroy `jobspec`.
     }
     return 0;
 }
@@ -203,7 +190,6 @@ static int depend_cb (flux_plugin_t *p,
 static void setup_rpc_cb (flux_future_t *f, void *arg)
 {
     int success = false;
-    json_t *env = NULL;
     const char *errstr = NULL;
     flux_t *h = flux_future_get_flux(f);
     struct create_arg_t *args = flux_future_aux_get (f, "flux::setup_args");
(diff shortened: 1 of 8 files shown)


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -53,33 +57,53 @@ static int raise_job_exception(flux_t *h, flux_jobid_t id, const char *exception
...
+                        " with errstr '%s'",
                         DWS_EPILOG_NAME,
-                        id);
+                        id,
+                        errstr);
         return -1;
     }
     if (!success) {
+        errno = EPROTO;
         flux_log_error (h, "Failed to clean up DWS workflow object for job %" PRIu64, id);
```
grondo, line 107: You could also switch `flux_log_error()` to flux `flux_log (h, LOG_ERR, "Failed....")`. Then `strerror (errno)` isn't automatically appended to the error message. This is common practice when an errno is unnecessary or doesn't make sense in the context of the error message.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -53,33 +57,53 @@ static int raise_job_exception(flux_t *h, flux_jobid_t id, const char *exception
...
             return -1;
         }
         *prolog_active = 0;
         if (!success) {
-            flux_log_error (h, "Failed to setup DWS workflow object for job %" PRIu64, id);
+            errno = EPROTO;
+            flux_log_error (h,
+                            "Failed to setup DWS workflow object for job %"
+                            PRIu64,
+                            id);
```
grondo, line 83: Same suggestion, perhaps `flux_log (h, LOG_ERR, ...` would be more appropriate here.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -34,7 +35,10 @@ struct create_arg_t {
 };
 
 
-static int raise_job_exception(flux_t *h, flux_jobid_t id, const char *exception, const char *errstr)
+static int raise_job_exception(flux_t *h,
+                               flux_jobid_t id,
+                               const char *exception,
+                               const char *errstr)
```
grondo, line 41: unrelated formatting changes? Seems ok for now since this is a small change, but as noted before it is usually better to keep these in a separate commit.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -488,24 +494,41 @@ static void prolog_remove_msg_cb (flux_t *h,
...
-        || flux_jobtap_job_aux_set (p, jobid, "flux::dws_run_started", (void *) 1, NULL) < 0){
-        dws_prolog_finish (h, p, jobid, 0, "failed to post dws_environment event", &prolog_active);
+        || flux_jobtap_job_aux_set (p,
+                                    jobid,
+                                    "flux::dws_run_started",
+                                    (void *) 1,
+                                    NULL) < 0) {
+        dws_prolog_finish (h, p, jobid, 0,
+                           "failed to post dws_environment event",
+                           prolog_active);
```
grondo, line 551: This commit was a little difficult to review because it seems to be mostly formatting changes, making it difficult to see what's actually going on. If it isn't too much trouble I'd suggest a separate formatting update commit, then the code changes will be more obvious. (However, if it is a pain, don't worry about this time)

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -188,7 +196,12 @@ static int depend_cb (flux_plugin_t *p,
         }
         *prolog_active = 0;
         flux_future_t *create_fut = flux_rpc_pack (
-            h, "dws.create", FLUX_NODEID_ANY, 0, "{s:O, s:I, s:O, s:i}", "dw_directives", dw, "jobid", id, "resources", resources, "userid", userid
+            h, "dws.create", FLUX_NODEID_ANY, 0,
+            "{s:O, s:I, s:O, s:i}",
+            "dw_directives", dw,
+            "jobid", id,
+            "resources", resources,
+            "userid", userid
```
grondo, line 202: More unrelated formatting changes?

grondo (approved):
Nice! The separate commits in this PR really made it easy to review, thanks!

Just a few minor suggestions inline, mainly regarding mixing actual code changes and formatting/whitespace updates, which can make it difficult to tell what is going on. However, your breakdown was done so well I'd consider it only a minor inconvenience this time.

Also, as an aside it seems like the line breaks in your commit messages might be set too small? Usually the commit title is broken at ~52 characters and commit body at 72 - it seemed like you might be wrapping the commit bodies at around 52? Not a big deal, just thought I'd point it out in case it was an editor setting.

I didn't see anything mandatory to fix, so approving!

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -462,6 +429,81 @@ static int exception_cb (flux_plugin_t *p,
...
+        return;
+    }
+    if (flux_jobtap_dependency_remove (p, jobid, CREATE_DEP_NAME) < 0) {
+        raise_job_exception (h, jobid, CREATE_DEP_NAME, "Failed to remove dependency for job");
+        return;
+    }
+    // if (json_object_set (jobspec, "resources", resources) < 0){
+    //     raise_job_exception (h, jobid, CREATE_DEP_NAME, "Failed to update jobspec with DWS resources");
+    //     goto done;
+    // }
```
cmoussa1, line 505: feel free to ignore me if this is on purpose but it looks like there is some commented-out code here

cmoussa1 (approved):
LGTM too! I agree with @grondo that this was a very well structured PR, great job! I noticed just one commented-out code block (but maybe that's intended so feel free to ignore if so).

garlick:
Just curious: would the problem with the RPC as it was before be addressed if pending RPCs were automatically sent an error response if the service disconnects?  If so then maybe there is a flux-core issue to open...

jameshcorbett:
> Just curious: would the problem with the RPC as it was before be addressed if pending RPCs were automatically sent an error response if the service disconnects? If so then maybe there is a flux-core issue to open...

That would be an improvement over the original, but I think this PR is an improvement over that. If RPCs automatically sent an error response, when the service crashed the jobtap plugin could raise an exception and remove the prolog/epilog, so jobs would never be stuck. But users' jobs would fail simply because the service crashed.

However, with this PR, the service can crash repeatedly and users' jobs will be unaffected.

jameshcorbett:
Thanks for the reviews @grondo and @cmoussa1 ! I addressed your formatting/whitespace comments (breaking them out into a separate PR), and will set MWP.

Issue #79, opened 2023-07-11
consider setting optional environment variables for Cray PMI


garlick:
In early design discussions for the `cray-pals` plugin, (summarized in #6), a modest list of environment variables was thought required to bootstrap libpals and in turn Cray's PMI library.   And indeed this seems to be mostly adequate.

In a recent issue,  a user was launching an application using Flux, and the application (linked with Cray MPICH) was wrapped in a singularity container.  It would segfault in `MPI_Init()` and a debugger trace led to code inside Cray `libpmi2.so` that was trying to establish peer connections.  Apparently it looks for an integer token in the PMI_SHARED_SECRET environment variable to authenticate peers, but if that is not set, it uses [munge](https://github.com/dun/munge) to authenticate.  Presumably munge was failing due to some aspect of running in a container and a segfault was the result.  It was thought that setting PMI_SHARED_SECRET would lead to a more robust `cray-pals` plugin.

In the course of that discussion, @RaymondMichael  provided the following note:

> Here’s a list of other variables that our launchers frequently set:
 
* PALS_APID - Unique application identifier
* PALS_APINFO - Path to libpals application information file
* PALS_DEPTH - The number of threads per process
* PALS_FD - File descriptor for libpals communication
* PALS_LOCAL_RANKID - This process is the N'th rank on this host
* PALS_LOCAL_SIZE - The total number of ranks on just this host
* PALS_NODEID - This process's node index
* PALS_RANKID - This process's rank index
* PALS_SPOOL_DIR - Local application spool directory for transferred files
* PMI_JOBID - The workload manager job ID
* PMI_LOCAL_RANK - This process is the N'th rank on this host
* PMI_LOCAL_SIZE - The total number of ranks on just this host
* PMI_RANK - This process's rank. PMI rank reordering may overrule this
* PMI_SHARED_SECRET - Key used to secure PMI communication
* PMI_SIZE - The total number of ranks
* PMI_UNIVERSE_SIZE - The maximum number of ranks for this application
* SLINGSHOT_DEVICES - The set of Slingshot devices configured for us
* SLINGSHOT_SVC_IDS - The Slingshot service ID for each corresponding device
* SLINGSHOT_TCS - The Slingshot traffic classes they can use
* SLINGSHOT_VNIS - The Slingshot security tokens they can use

[...]

PR #84, opened 2023-07-25
Cray pals env vars

jameshcorbett:
This is a partial resolution of #79. The only new variable set is `PMI_SHARED_SECRET`. Some additional variables are unset.

We could set more variables listed in #79, but only `PMI_SHARED_SECRET` has resulted in an error.

### DIFF
--- a/src/job-manager/plugins/cray_pals_port_distributor.c
+++ b/src/job-manager/plugins/cray_pals_port_distributor.c
@@ -92,8 +92,10 @@ static void count_job_shells (flux_future_t *fut, void *arg)
     json_int_t port1, port2;
     flux_t *h;
     json_t *arr = NULL;
+    int prolog_status = 0;
 
     if (!(h = flux_future_get_flux (fut))) {
+        prolog_status = 1;
         goto cleanup;
     }
     if (flux_kvs_lookup_get_unpack (fut,
@@ -105,6 +107,7 @@ static void count_job_shells (flux_future_t *fut, void *arg)
         flux_log_error (h,
                         PLUGIN_NAME ": "
                         "Error fetching R from shell-counting future");
+        prolog_status = 1;
         goto cleanup;
     }
     if (hostlist_count (hlist) == 1) {
@@ -126,6 +129,7 @@ static void count_job_shells (flux_future_t *fut, void *arg)
                                     (flux_free_f)json_decref) < 0) {
         if (arr)
             json_decref (arr);
+        prolog_status = 1;
         flux_log_error (h,
                         PLUGIN_NAME ": "
                         "Failed to post ports to job");
@@ -136,7 +140,7 @@ cleanup:
     if (flux_jobtap_prolog_finish (triple->plugin,
                                    triple->jobid,
                                    "cray-pals-port-distributor",
-                                   0) < 0)
+                                   prolog_status) < 0)
         flux_log_error (h,
                         PLUGIN_NAME ": prolog_finish");
     if (hlist)
--- a/src/job-manager/plugins/cray_pals_port_distributor.c
+++ b/src/job-manager/plugins/cray_pals_port_distributor.c
@@ -14,8 +14,10 @@
 
 #include <stdio.h>
 #include <syslog.h>
+#include <stdint.h>
 
 #include <jansson.h>
+#include <sodium.h>
 
 #include <flux/core.h>
 #include <flux/hostlist.h>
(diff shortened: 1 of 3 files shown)


src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -269,6 +275,10 @@ int flux_plugin_init (flux_plugin_t *p)
                   port_max);
     }
 
+    if (sodium_init() == -1){
+        flux_log_error (h, "error initializing libsodium");
+        return -1;
+    }
```
garlick, line 281: I don't think `sodium_init()` sets errno?  If not, `flux_log (h, LOG_ERR, ...)` would be better here.

src/shell/plugins/cray_pals.c
```diff
@@ -624,6 +625,14 @@ static int read_future (flux_future_t *fut, char *buf, size_t bufsize)
...
                 }
             }
+            if (!(random_entry = json_object_get (context, "random_integer"))
+                || !(json_is_integer (random_entry))) {
+                json_decref (o);
+                shell_log_error ("Error fetching random integer");
+                return -1;
+            } else {
+                *random = json_integer_value (random_entry);
+            }
```
garlick, line 635: This block could be simplified to
```c
if (json_unpack (context, "{s:I}", "random_integer", random) < 0) {
    ... handle error
}
```
It could actually be combined in one `json_unpack()` with the parsing of `ports` above it but maybe that's a future cleanup.

src/shell/plugins/cray_pals.c
```diff
@@ -643,21 +652,24 @@ static int get_pals_ports (flux_shell_t *shell, json_int_t jobid)
...
     /* read_future() returns 1 if port distribution event was found:
      */
-    if (rc == 1
-        && flux_shell_setenvf (shell, 1, "PMI_CONTROL_PORT", "%s", buf) < 0) {
-        return -1;
+    if (rc == 1){
+        if (flux_shell_setenvf (shell, 1, "PMI_CONTROL_PORT", "%s", buf) < 0
+            || flux_shell_setenvf (shell, 1, "PMI_SHARED_SECRET", "%" JSON_INTEGER_FORMAT, random)) {
+            return -1;
+        }
```
garlick, line 672: Suggest using `%ju` format here and casting random to `uintmax_t` so that if the earlier suggestion is taken to use 64 bits of random key, the value isn't ever formatted as a negative number.

garlick (commented):
A couple of comments for you.

jameshcorbett:
Thanks @garlick , I addressed all your comments!

src/shell/plugins/cray_pals.c
```diff
@@ -643,21 +643,24 @@ static int get_pals_ports (flux_shell_t *shell, json_int_t jobid)
...
     flux_future_destroy (fut);
 
     /* read_future() returns 1 if port distribution event was found:
      */
-    if (rc == 1
-        && flux_shell_setenvf (shell, 1, "PMI_CONTROL_PORT", "%s", buf) < 0) {
-        return -1;
+    if (rc == 1){
+        if (flux_shell_setenvf (shell, 1, "PMI_CONTROL_PORT", "%s", buf) < 0
+            || flux_shell_setenvf (shell, 1, "PMI_SHARED_SECRET", "%ju", (uintmax_t) random)) {
```
garlick, line 661: Probably should check for `< 0` for the error case like the line above it (even though I think checking for `true` also happens to work here)

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -113,15 +116,18 @@ static void count_job_shells (flux_future_t *fut, void *arg)
     if (hostlist_count (hlist) == 1) {
         goto cleanup;  // no need to post ports
     }
+    randombytes_buf(&random, sizeof (json_int_t));
```
garlick, line 119: style: space between function name and open paren

src/job-manager/plugins/cray_pals_port_distributor.c
```diff
@@ -269,6 +275,10 @@ int flux_plugin_init (flux_plugin_t *p)
                   port_max);
     }
 
+    if (sodium_init() == -1){
```
garlick, line 278: style: space between function name and open paren; space between conditioanal close paren and open brace.

jameshcorbett:
Good catches, fixed. Setting MWP

PR #87, opened 2023-08-07
dws: improve error message

### LINKED ISSUE #86: confusing job exception note when dws service is not loaded
On tioga, the flux-coral2-dws systemd service failed to load after a Flux restart. The issue reported by a user was unclear though:
```
0.045s: job.exception type=dws-create severity=0 DWS workflow interactions failed: Failed to unpack dws.create RPC
```

In this code:

https://github.com/flux-framework/flux-coral2/blob/a543de95bc4a370f2e4ef45a9e7fe9f1beac79bc/src/job-manager/plugins/dws-jobtap.c#L125-L132

An `ENOSYS` error could be returned if the dws service isn't loaded, but the code always assumes an unpack error.
Maybe just adding `strerror (errno)` would be sufficient to call out that the service is not loaded.

jameshcorbett:
Problem: the error message given when the dws
service is not available is misleading and
unhelpful.

Improve the error message.

Fixes #86.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -16,6 +16,7 @@
 #include "config.h"
 #endif
 
+#include <errno.h>
 #include <stdio.h>
 #include <syslog.h>
 #include <stdint.h>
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

PR #88, opened 2023-08-18
Build a resource graph with "rabbit" under "rack"

jameshcorbett:
As a first pass at getting rabbits through sched, @trws and I discussed making "rabbit" vertices contained by "rack" vertices. "rack" vertices will contain all the nodes with PCIe connections to the rabbit. This will allow jobspecs like

```json
  "resources": [
    {
      "type": "rack",
      "count": 1,
      "with": [
        {
          "type": "node",
          "count": 1,
          "exclusive": true,
          "with": [
            {
              "type": "slot",
              "count": 1,
              "with": [
                {
                  "type": "core",
                  "count": 1
                }
              ],
              "label": "task"
            }
          ]
        },
        {
          "type": "rabbit",
          "count": 1,
          "with": [
            {
              "type": "ssd",
              "count": 2048,
              "exclusive": true
            }
          ]
        }
      ]
    }
  ]
```

The dws2jgf script is meant to generate the resource graph for rabbit systems. It takes in system R and adds the graph under the "scheduling" key. 

The script has some problems--it mixes up broker ranks, drops properties, and makes assumptions about homogeneity of compute nodes. But the resource graph it builds has the right structure, and it's an improvement over the version of the script that exists in master.


jameshcorbett:
Force-pushed to clean up the history a bit

t/t2000-dws2jgf.t
```diff
@@ -4,19 +4,65 @@ test_description='Test dws2jgf command'
 
 . $(dirname $0)/sharness.sh
 
+FLUX_SIZE=2
+
+test_under_flux ${FLUX_SIZE} job
+
+flux setattr log-stderr-level 1
+
 CMD=${FLUX_SOURCE_DIR}/src/cmd/flux-dws2jgf.py
```
grondo, line 13: Not necessary to change anything, but I just wanted to point out that if you append `${FLUX_SOURCE_DIR}/src/cmd` to `FLUX_EXEC_PATH`, then you should be able to call this script as `flux dws2jgf` instead of `flux python ${CMD}`...

grondo (approved):
I'm not qualified to review most of this PR, but  what I could review looked good to me. Nice job organizing the commits. I put one note inline just as an fyi.

PR #89, opened 2023-08-21
Jobspec manipulations

jameshcorbett:
The first step towards making resource requests that match the resource graph built in #88 is to create jobspecs with top-level "rack" resources. This PR adds logic for building such jobspecs.


cmoussa1 (approved):
LGTM!

PR #90, opened 2023-08-23
DWS resource graph improvements

jameshcorbett:
Fixing some of the issues with the resource graph introduced by #88 .


t/data/dws2jgf/expected-compute-01-04.jgf
```diff
@@ -1051,7 +1051,7 @@
             "type": "ssd",
             "basename": "ssd",
             "name": "ssd2",
-            "id": 1,
+            "id": 2,
```
cmoussa1, line 1054: not sure if this is intended, but it looks like there's a jump from 0 to 2 here instead of sequential like the rest of the file (i.e it goes from `ssd0` to `ssd2`); this seems like a large manual change, though, so if it's no big deal to leave it like this, that's fine with me!

cmoussa1 (approved):
LGTM! Just had one minor comment about some numbering in the test `.jgf` files. Thanks @jameshcorbett!

src/cmd/flux-dws2jgf.py
```diff
@@ -193,6 +194,7 @@ def _encode(self):
 
 
 def get_node_children(r_lite):
+    """Return a mapping from rank to children (cpus, gpus, etc.)"""
```
grondo, line 196: This added line probably belongs in a different commit?

t/data/dws2jgf/expected-compute-01-04.jgf
jameshcorbett: Good catch! I think this is just the diff being misleading though. The actual file introduced by the PR has ssd0, ssd1, ssd2, and so on. They used to be all called "ssd1" though, so "ssd1" doesn't show up in the diff.

src/cmd/flux-dws2jgf.py
```diff
@@ -206,6 +208,25 @@ def get_node_children(r_lite):
     return rank_to_children
 
 
+def get_node_properties(properties):
```
grondo, line 210: You are probably aware of this, but flux-core exports a [ResourceSet](https://github.com/flux-framework/flux-core/blob/master/src/bindings/python/flux/resource/ResourceSet.py)  API which may help reduce some of the code here. For example, theres a `get_properties()` method that returns a `properties` object in [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html) format, which is literally a mapping of properties to rank idsets.

Not necessary to change this working solution, but just thought I'd mention it.

grondo (approved):
LGTM!  Just a couple minor comments inline.

src/cmd/flux-dws2jgf.py
jameshcorbett: Good catch, will fix, thanks.

src/cmd/flux-dws2jgf.py
jameshcorbett: I didn't know that, actually. I'll take a look and rework this.

grondo:
One other trivial comment is that the lines in the commit messages (body part) seem really short. The standard is to break at column 72, which leaves the lines short enough to not wrap in `git log` but not so short that it is a struggle to read. No need to change anything though if that is your preference, though!

jameshcorbett:
> One other trivial comment is that the lines in the commit messages (body part) seem really short. The standard is to break at column 72, which leaves the lines short enough to not wrap in `git log` but not so short that it is a struggle to read. No need to change anything though if that is your preference, though!

I'll fix that. I write my commits in the terminal, so I have no idea how many characters I've written, I just eyeball it... would probably save myself some effort if I would just write them in an editor.

src/cmd/flux-dws2jgf.py
jameshcorbett: Actually I'm not sure using the ResourceSet would save much complexity, because the way it's structured now, I want a mapping from ranks to properties, not the other way around. So I think I'll keep it the way it is for now.

jameshcorbett:
OK, I addressed all your comments @grondo .

grondo:
Great, still LGTM! :-)

t/data/dws2jgf/expected-compute-01-04.jgf
cmoussa1: ohhhh, i see - sorry for the confusing suggestion!! you can ignore me then 😆

PR #91, opened 2023-08-23
mergify: remove status check requirement

jameshcorbett:
Problem: mergify is configured to require status
checks to pass before merging, but we can use
GitHub's branch protections instead.

Remove mergify's status checks requirements.


grondo (commented):
Great, looks good!

grondo:
Oh, I'll press the merge button since mergify config was touched.

PR #92, opened 2023-08-25
Fluxion scheduling for rabbits

jameshcorbett:
The final piece after #88, #89, and #90 for getting rabbit allocations through Fluxion. The logic for building a resource graph with rabbits is there already, and so is the logic for creating jobspecs which request rabbits. But at the moment the scheduler never actually gets to see the jobspecs that request rabbits, because the plugins hold them back. This PR fixes that.

It also fixes (or at least minimizes) a major limitation of the previous implementation's handling of lustre file systems, which stuck each lustre allocation on a single rabbit. Now lustre file systems are split across rabbits by default.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -168,12 +168,13 @@ static int depend_cb (flux_plugin_t *p,
     json_t *dw = NULL;
     flux_t *h = flux_jobtap_get_flux (p);
     json_t *resources;
+    json_t *jobspec;
     int userid;
     int *prolog_active = NULL;
 
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:I s:{s:{s:{s?o}}} s:{s:o} s:i}",
+                                "{s:I s:{s:{s:{s?o}}} s:{s:o} s:i s:o}",
                                 "id",
                                 &id,
                                 "jobspec",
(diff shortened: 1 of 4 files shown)


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -560,12 +574,20 @@ static void resource_update_msg_cb (flux_t *h,
...
     json_t *resources = NULL;
+    json_t *jobspec;
 
     if (flux_msg_unpack (msg, "{s:I, s:o}", "id", &jobid, "resources", &resources)
         < 0) {
         flux_log_error (h, "received malformed dws.resource-update RPC");
         return;
     }
+    if (!(jobspec = flux_jobtap_job_aux_get (p, jobid, "dws_jobspec"))
+        || json_object_set (jobspec, "resources", resources) < 0) {
```
grondo, line 585: This should probably be done via an RFC 21 [jobspec-update](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#jobspec-update-event). I think that would simplify the implementation here too. However, these events are currently restricted to updating keys under `attributes.`. I'm not sure why that is though, so I'll open a issue in the RFC repo and get this sorted out. 

For now this will work, but we'd have to analyze if jobspec updates will be lost for pending jobs on a job manager restart (I haven't looked, it may well work, but was just bringing this up as a possible concern)

grondo (approved):
LGTM! A note inline about possibly switching to a `jobspec-update` event for jobspec modifications in the future, but that isn't possible now due to limitations specified in RFC 21.

src/job-manager/plugins/dws-jobtap.c
grondo: FYI: https://github.com/flux-framework/rfc/issues/395

jameshcorbett: Thanks for the info @grondo !

PR #96, opened 2023-09-11
testsuite: load content module in rc scripts

garlick:
Problem: flux-core is migrating the content service from the broker to a broker module.

Load the content module in test rc scripts.

Ignore any errors from 'flux module load' so that flux-coral2 continues to work with older flux-core versions.


jameshcorbett (approved):
Looks reasonable to me!

PR #97, opened 2023-09-11
rc: call set -e after shebang in 01-coral2-rc

### LINKED ISSUE #95: rc: avoid `#!/bin/bash -e` in rc scripts
As noted in flux-framework/flux-core#5447, use of `#!/bin/bash -e` in Flux rc scrips can break for users that set `BASH_ENV=/etc/bashrc` since `bashrc` is not "errexit clean"

The proposed solution is to drop `-e` for `set -e` after the she-bang.

garlick:
Problem: If BASH_ENV is set to a script that is not errexit clean, the 01-coral2-rc script may immediately.

Drop -e from the shebang and add a "set -e" afterwards.

Fixes #95

### DIFF
--- a/etc/01-coral2-rc
+++ b/etc/01-coral2-rc
@@ -1,4 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
+
+set -e
 
 # Load jobtap plugin, unless it's already loaded (e.g. by config)
 jobtap_load() {


jameshcorbett (approved):
LGTM, thanks. Not sure why the tests failed, restarting...

jameshcorbett:
@garlick I rebased to see if that fixes the test failures.

PR #98, opened 2023-09-12
rc: use posix shell

### LINKED ISSUE #95: rc: avoid `#!/bin/bash -e` in rc scripts

garlick:
Problem: 01-coral2-rc has a bash shebang, but when a user sets BASH_ENV, it potentially sources unnecessary stuff.

Switch to a posix shebang.
Re-add the -e since this also fixes #95.

Sorry I shouldn't have let that last PR go in.  The solution kind of changed to this better one right about then.

### DIFF
--- a/etc/01-coral2-rc
+++ b/etc/01-coral2-rc
@@ -1,6 +1,4 @@
-#!/bin/bash
-
-set -e
+#!/bin/sh -e
 
 # Load jobtap plugin, unless it's already loaded (e.g. by config)
 jobtap_load() {


jameshcorbett:
> Switch to a posix shebang. Re-add the -e since this also fixes #95.

I don't see a re-added `-e` though...?

> Sorry I shouldn't have let that last PR go in. The solution kind of changed to this better one right about then.

No worries, that was my bad.

garlick:
> I don't see a re-added -e though...?

Doh!  Just re-pushed.

jameshcorbett (approved):
LGTM! I'll let you set MWP this time around ;)

PR #102, opened 2023-10-04
Fix dws-jobtap hang

### LINKED ISSUE #101: dws-jobtap: RPC futures could be leaked if plugin unloaded before responses
The futures resulting from RPCs in the dws-jobtap plugin are only destroyed in their respective callbacks and therefore may be leaked if the plugin is unloaded or the broker stops before any responses. This can cause a hang at exit as seen in flux-framework/flux-core#5484.

One solution would be to tie destruction of these futures to the job with `flux_jobtap_job_aux_set()`. Another solution would be to collect outstanding futures in a list and destroy them all on plugin exit.

jameshcorbett:
Fixes #101 .


grondo (approved):
Nice work! LGTM!

PR #103, opened 2023-10-05
NEWS: add 0.8.0 release notes

jameshcorbett:
Problem: there are no release notes for 0.8.0.

Add them.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,16 @@
+flux-coral2 0.8.0 - 2023-10-04
+------------------------------
+
+### Fixes
+
+ * dws: update group of k8s crds (#101)
+ * rc: use posix shell (#98)
+
+### Testsuite
+
+ * testsuite: remove usage of 'flux mini' (#102)
+ * testsuite: load content module in rc scripts (#96)
+
 flux-coral2 0.7.0 - 2023-08-30
 ------------------------------
 


NEWS.md
```diff
@@ -1,3 +1,16 @@
+flux-coral2 0.8.0 - 2023-10-04
+------------------------------
+
+### Fixes
+
+ * dws: update group of k8s crds (#101)
+ * rc: use posix shell (#98)
+
```
grondo, line 8: Are these missing here?
```
 * Fix dws-jobtap hang (#102)
 * rc: call set -e after shebang in 01-coral2-rc (#97)
```
Oh I see #102 is labeled as `remove usage of 'flux-mini'` but it could be more useful to call out both changes?

BTW, we have a useful script in flux-core tree `src/test/relnotes.sh` that print out all merged PRs since the last tag, e.g.:
```
$ ../flux-core.git/src/test/relnotes.sh 
Changes since v0.7.0:
 * dws: fix group of k8s crds (#100)
 * Fix dws-jobtap hang (#102)
 * rc: use posix shell (#98)
 * rc: call set -e after shebang in 01-coral2-rc (#97)
 * testsuite: load content module in rc scripts (#96)
```

Not sure if that's helpful here or not.

grondo (approved):
LGTM! I did note that the release notes here don't seem complete, but I'll leave that to you to decide if and how to rectify.

NEWS.md
jameshcorbett: Ack that's what I get for setting MWP before getting a review--won't do that again. I think #97 was superseded by #98 so it's just #102 that's missing. But I guess I'll leave it as-is.

jameshcorbett:
Thanks @grondo !

Issue #104, opened 2023-10-09
Distributing rabbit hostnames as environment variables


jameshcorbett:
Marty McFadden asked whether Flux could export the hostname of the rabbit local to each compute node in a job. He wants to use the functionality described [here](https://nearnodeflash.github.io/v0.0.6/guides/user-containers/readme/#container-spec).

> I would like for the SCR library on each compute node to be able to open a socket to the available port on its local rabbit node.

The Flux-DWS plugins already export some DWS-specified environment variables. The variables are posted in an [eventlog entry](https://github.com/flux-framework/flux-coral2/blob/71818bfb7516aa22b7062fec28c007cd8c072ff4/src/job-manager/plugins/dws-jobtap.c#L639). I was thinking I could add a mapping from rabbit hostname to compute node hostlist to the event context, and let the job shell plugin set the right environment variable?

grondo:
In case you were asking for feedback, I can't think of a much better way to do this than you've outline above.

My only idea is you could map the rabbit hostnames to job shell ranks (i.e. node ranks relative to the job) and just send set a single hostlist in the eventlog entry. This would make it easier for the job shell plugin, it could just render the hostname from the hostlist based on its rank in `flux_shell_info(3)` (the job shell rank).

jameshcorbett:
I was asking for feedback actually, but even if I wasn't it's always helpful, so thanks! I think it will be easier in other places to send a mapping of hostlists though, since `cora2_dws.py` (which has the mapping from rabbits to compute nodes) won't always have the job shell ranks available.

PR #105, opened 2023-10-10
Rabbit env vars

jameshcorbett:
As discussed in #104, export the hostname of the rabbit local to each compute node as the FLUX_LOCAL_RABBIT environment variable.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -618,10 +618,10 @@ static void prolog_remove_msg_cb (flux_t *h,
 {
     flux_plugin_t *p = (flux_plugin_t *)arg;
     json_int_t jobid;
-    json_t *env = NULL;
+    json_t *env = NULL, *rabbit_mapping = NULL;
     int *prolog_active, junk_prolog_active = 1;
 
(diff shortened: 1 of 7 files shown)


grondo (approved):
This looks good to me!

My only suggestion would be to add a test with multiple rabbits with multiple nodes to further exercise selection of the correct rabbit hostname in the shell plugin. I'd call that optional though and leave it to you whether to implement the extra testing.

grondo:
One other worry could be the size of the eventlog entry when there are many rabbits and many nodes. I don't think there will be a problem there, but we might want to test the largest entry we expect to see just in case.

jameshcorbett:
> This looks good to me!
> 
> My only suggestion would be to add a test with multiple rabbits with multiple nodes to further exercise selection of the correct rabbit hostname in the shell plugin. I'd call that optional though and leave it to you whether to implement the extra testing.

Force-pushed to add another test. I'll set MWP.

jameshcorbett:
> One other worry could be the size of the eventlog entry when there are many rabbits and many nodes. I don't think there will be a problem there, but we might want to test the largest entry we expect to see just in case.

I think all the rabbits will be attached to nodes with contiguous IDs, so the hostlists should be relatively compact, along the lines of `elcapitan[0-18]` rather than `elcapitan[201,302,305,307,...]`. But yeah maybe a test with a thousand rabbits or something in that ballpark would be good.

PR #110, opened 2023-11-03
Formalize libsodium dependency

### LINKED ISSUE #108: Missing dependency on libsodium
cray_pals_port_distributor.c uses functions from libsodium (specifically `randombytes_buf`) but the Makefile does not pull in the sodium library. This leads to undefined symbols at runtime. Maybe this previously came from flux-core's dependency on libsodium; I'm not sure, I haven't tracked it down.

jameshcorbett:
Fixes #108 .


src/job-manager/plugins/Makefile.am
```diff
@@ -1,7 +1,8 @@
 AM_CFLAGS = \
 	$(WARNING_CFLAGS) \
 	$(CODE_COVERAGE_CFLAGS) \
-    $(FLUX_CORE_CFLAGS)
+    $(FLUX_CORE_CFLAGS) \
+    $(SODIUM_CFLAGS)
```
grondo, line 5: Should `$(JANSSON_CFLAGS)` be included in `AM_CFLAGS` as well? This would be required, for example, if jansson were installed in a non-standard path. However, I'm not sure if this is covered some other way, so ignore me if this is fine.

jameshcorbett: I don't believe it's covered any other way, let me force-push with that added.

garlick (approved):
Looks good - and I just confirmed that configure fails if libsodium is not installed.
There's a little whitespace incongruity evident in the github diff rendering in shell/plugins/Makefile.am but that seems pretty low priority and could be fixed later.

jameshcorbett:
Thanks! Setting MWP.

PR #114, opened 2023-11-30
dws: disable fluxion scheduling of rabbits

jameshcorbett:
Problem: fluxion scheduling of rabbits is working but requires some configuration changes on the machines it would be used on (LC's El Cap EA machines).

Disable fluxion scheduling temporarily until the config changes can be worked out.

This commit should be reverted before the next major release.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -341,9 +341,7 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
             "job-manager.dws.resource-update",
             payload={
                 "id": jobid,
-                "resources": directivebreakdown.apply_breakdowns(
-                    k8s_api, workflow, resources
-                ),
+                "resources": resources,
             },
         )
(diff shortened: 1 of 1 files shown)


grondo (commented):
Ok, I don't really understand the change here (unsurprising), but everything LGTM.

One alternative might be to somehow add a configuration parameter and default it to off so you don't have to revert a commit later (just update configuration when the support is ready).

grondo (approved):
Meant to approve.

jameshcorbett:
> Ok, I don't really understand the change here (unsurprising), but everything LGTM.
> 
> One alternative might be to somehow add a configuration parameter and default it to off so you don't have to revert a commit later (just update configuration when the support is ready).

Do you mean have something like a command-line argument, and then add a check like `if ARGUMENT: do thing`? That makes sense but I would want to remove that check eventually so I don't really think it would save any git churn.

grondo:
Sounds reasonable!

PR #116, opened 2023-11-30
README: fix broken links

jameshcorbett:
It seems like github tries to send the `github.com` links to someplace within this repository, rather than to an external repository. Adding `https://` seems to fix it.


grondo (approved):
👍

PR #117, opened 2023-12-06
Draining rabbits

jameshcorbett:
Leverages functionality provided by [this flux-sched PR](https://github.com/flux-framework/flux-sched/pull/1110) to mark rabbits up or down based on status reported by Kubernetes 'Storage' objects.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -623,7 +623,6 @@ def main():
         raise
     populate_rabbits_dict(k8s_api)
     handle = flux.Flux()
-    services = register_services(handle, k8s_api)
     # create a timer watcher for killing workflows that have been stuck in
     # the "Error" state for too long
     handle.timer_watcher_create(
@@ -636,6 +635,7 @@ def main():
     # or new RPCs are received
     with Watchers(handle, watch_interval=args.watch_interval) as watchers:
         init_rabbits(k8s_api, handle, watchers, args.resourcegraph)
+        services = register_services(handle, k8s_api)
         watchers.add_watch(
             Watch(k8s_api, WORKFLOW_CRD, 0, workflow_state_change_cb, handle, k8s_api)
         )
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -16,6 +16,7 @@ import argparse
 import logging
 import pwd
 import time
+import pathlib
 
 import kubernetes as k8s
 from kubernetes.client.rest import ApiException
@@ -507,6 +508,13 @@ def setup_parsing():
         default=None,
         metavar="FILE",
         help="Path to kubeconfig file to use",
+    )
+    parser.add_argument(
+        "--resourcegraph",
+        "-r",
+        default=str(pathlib.Path("/etc/flux/system/R").absolute()),
+        metavar="FILE",
+        help="Path to file containing Fluxion JGF resource graph",
     )
     return parser
 
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -56,6 +56,16 @@ class WorkflowInfo:
         self.rabbits = None
 
 
+def log_rpc_response(rpc):
+    """RPC callback for logging response."""
+    try:
+        msg = rpc.get()
+    except Exception as exc:
+        LOGGER.warning("RPC error %s", str(exc))
+    else:
+        LOGGER.debug("RPC response was %s", msg)
+
+
 def fetch_rabbits(k8s_api, workflow_computes):
     """Fetch all the rabbits associated with this workflow"""
     response = k8s_api.get_namespaced_custom_object(
@@ -155,7 +165,7 @@ def create_cb(handle, _t, msg, api_instance):
     handle.rpc(
         "job-manager.memo",
         payload={"id": jobid, "memo": {"rabbit_workflow": workflow_name}},
-    )
+    ).then(log_rpc_response)
 
 
 @message_callback_wrapper
@@ -180,7 +190,7 @@ def setup_cb(handle, _t, msg, k8s_api):
     handle.rpc(
         "job-manager.memo",
         payload={"id": jobid, "memo": {"rabbits": list(nodes_per_nnf.keys())}},
-    )
+    ).then(log_rpc_response)
     k8s_api.patch_namespaced_custom_object(
         COMPUTE_CRD.group,
         COMPUTE_CRD.version,
@@ -325,7 +335,9 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
         # delete workflow object and tell DWS jobtap plugin that the job is done
         k8s_api.delete_namespaced_custom_object(*WORKFLOW_CRD, winfo.name)
         winfo.deleted = True
-        handle.rpc("job-manager.dws.epilog-remove", payload={"id": jobid})
+        handle.rpc("job-manager.dws.epilog-remove", payload={"id": jobid}).then(
+            log_rpc_response
+        )
     elif winfo.toredown:
         # in the event of an exception, the workflow will skip to 'teardown'.
         # Without this early 'return', this function may try to
@@ -344,7 +356,7 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
                 "id": jobid,
                 "resources": resources,
             },
-        )
+        ).then(log_rpc_response)
     elif state_complete(workflow, "Setup"):
         # move workflow to next stage, DataIn
         move_workflow_desiredstate(winfo.name, "DataIn", k8s_api)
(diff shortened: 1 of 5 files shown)


garlick:
Haven't been following this but just a quick check - if flux restarts, does the scheduler recover the state of the rabbits?

jameshcorbett:
Whenever the flux-coral2-dws service restarts, it polls all the rabbits and marks them each as up or down. Expensive, but I thought it was better that than the alternative. So I think that means the answer to your question is yes?

jameshcorbett:
Yeah, basically. Sounds like [RFC 28](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_28.html) all over again, only for rabbits instead of execution targets. In which case a streaming RPC would be appropriate.

One problem is that `flux-coral2-dws` can go down independently of Fluxion, so it would stop sending responses to the streaming RPC and Fluxion would have to send another.

I could maybe add some functionality to Fluxion so that whenever it comes online, it asks `flux-coral2-dws` for the status of all the rabbits. Combine that with what's in this PR, and I think Fluxion would always be in a good state. If `flux-coral2-dws` is down when Fluxion restarts, that's fine, all rabbit jobs will be rejected and Fluxion will get the status of the rabbits as soon as `flux-coral2-dws` comes back. If Fluxion is down when `flux-coral2-dws` restarts, well, I'm less certain what happens then to rabbit jobs, but at least Fluxion will ask for current rabbit status whenever it comes back.

garlick:
> I didn't know that Fluxion was restarted ever.

One thing I was not remembering when I commented was that the `flux-coral2-dws` systemd unit file contains
```
BindsTo=flux.service
After=flux.service
```
https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#BindsTo=

which I think would ensure that if flux crashes or is shut down, dws is stopped, and that dws is restarted only after flux is back up and has loaded the scheduler.

That leaves the case of manually reloading the scheduler on a live system.  This could come in handy, for example if we need to reconfigure or update the scheduler without affecting running jobs.  However it's less likely that than the above situations which are a regular certainty.

I didn't mean to get you off track.  Maybe we could open issue(s) on any loose ends here?  We did take pains to make it possible to reload the scheduler so it would be good to not give that capability up on El Cap.

grondo:
> Edit: could the scheduler allocate down rabbits to jobs right after it starts up but before it gets an RPC from dws?

This was going to be my question.

Also, I assume systems with rabbits are using Rv1 format and not `rv1_nosched`, otherwise the scheduler cannot be restarted due to #991.

jameshcorbett:
> > Edit: could the scheduler allocate down rabbits to jobs right after it starts up but before it gets an RPC from dws?
> 
> This was going to be my question.

I think the answer is no, because DWS sends the rabbit status RPCs before it starts receiving RPCs from the DWS jobtap plugin, which would mean that rabbit jobs would just be rejected in that time. 

Or at least, if it isn’t that way already, I can easily make it so by moving a line of code or two. I’ll verify when I get home from the quarterly.

jameshcorbett:
I opened an issue #118 to discuss more about Fluxion restarts. 

> Also, I assume systems with rabbits are using Rv1 format and not `rv1_nosched`, otherwise the scheduler cannot be restarted due to #991.

I guess so! At some point `flux-coral2-dws` may require the extra information emitted by Fluxion anyway.


> > > Edit: could the scheduler allocate down rabbits to jobs right after it starts up but before it gets an RPC from dws?
> > 
> > 
> > This was going to be my question.
> 
> I think the answer is no, because DWS sends the rabbit status RPCs before it starts receiving RPCs from the DWS jobtap plugin, which would mean that rabbit jobs would just be rejected (before they make it to the scheduler) in that time.
> 
> Or at least, if it isn’t that way already, I can easily make it so by moving a line of code or two. I’ll verify when I get home from the quarterly.

Pushed a new commit that does this. @grondo , I think it's ready for review now.

garlick:
The race I (and maybe @grondo) were thinking of would occur if
- rabbit resources are initially marked up
- the scheduler starts handling alloc requests
- the first RPC from DWS comes in that marks some of them down

It would seem like the right solution would be to mark all resources down initially and wait for something to come along and mark them up, however nothing would do that in a sub-instance.

It's possible I'm confused about how things work!

jameshcorbett:
Yeah I think I had the same race in mind. But I think the race is less likely given this consideration--if `coral2_dws.py` isn't running, any incoming job that requests rabbits will be rejected in the `DEPEND` state by the `dws_jobtap` plugin, because [the jobtap plugin requires the `dws` service](https://github.com/flux-framework/flux-coral2/blob/23b83fd67b9916954a1c6a489d8ea26c6851ffd5/src/job-manager/plugins/dws-jobtap.c#L134) which `coral2_dws` provides. If or when `coral2_dws` starts up, it will mark rabbits as down before registering the `dws` service and therefore rabbits will be marked down before new rabbit jobs are accepted by the `dws-jobtap` plugin. This is what I changed in [this commit](https://github.com/flux-framework/flux-coral2/pull/117/commits/e61859efd76bfb7f56ecab37eac268ff4df309d8).

So in the case where `coral2_dws` is _not_ running when the scheduler starts handling alloc requests, I don't think it could receive any alloc requests for _new_ rabbit jobs (i.e. submitted after the restart). However, it could receive alloc requests for jobs that had already made it past DEPEND when the scheduler restarted--they would (I think) sneak past the `dws_jobtap` plugin. This could be pretty bad I guess, thoughts?

That still leaves the case where `coral2_dws` is already running when the scheduler restarts, i.e. a scheduler restart on a live system. But that's a discussion for #118 I think.

In general though even though there are some race conditions on scheduler startup I think this PR is still a major improvement over the way things work now, in which rabbits are _always_ up. 

> It would seem like the right solution would be to mark all resources down initially and wait for something to come along and mark them up, however nothing would do that in a sub-instance.

That sounds great if it could be done in the config somehow, like in the Fluxion JGF. But I don't think that's possible at the moment...

grondo:
> It would seem like the right solution would be to mark all resources down initially and wait for something to come along and mark them up

This should be how things already work as part of the acquisition protocol?

e.g. see: https://github.com/flux-framework/flux-sched/issues/662#issuecomment-638892848

jameshcorbett:
> > It would seem like the right solution would be to mark all resources down initially and wait for something to come along and mark them up
> 
> This should be how things already work as part of the acquisition protocol?
> 
> e.g. see: [flux-framework/flux-sched#662 (comment)](https://github.com/flux-framework/flux-sched/issues/662#issuecomment-638892848)

That doesn't seem to mesh with my experience of initializing Fluxion with rabbits. I'm puzzled... because core doesn't know anything about them, it can't mark them as up, but I can still have Fluxion schedule against them. I wonder if it's because Fluxion is informed about its resources (and the rabbits) by reading a file which describes the graph, rather than some other way?

grondo:
Oh, yeah, sorry I was just making a point that the scheduler should assume "resources" are down until something tells them the are up. Since rabbits are not "execution targets" they do not fall under the purview of the resource acquisition protocol, but I was wondering if the same scheme should apply here...

jameshcorbett:
I made issue #119 for initializing rabbits as down. I have a path forward for implementing it.

src/modules/coral2_dws.py
```diff
@@ -401,58 +401,77 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
...
-    obj = event["object"]
-    name = obj["metadata"]["name"]
-    capacity = obj["status"]["capacity"]
-    status = obj["status"]["status"]
+def mark_rabbit(handle, status, resource_path):
+    """Send an RPC to mark a rabbit as up or down."""
+    if status == "Ready":
+        LOGGER.debug("Marking rabbit %s as up", resource_path)
+        payload = {"resource_path": resource_path, "status": "up"}
+        handle.rpc("sched-fluxion-resource.set_status", payload)
```
grondo, line 409: It appears this RPC could fail. I'm not sure what could be done about it, but should a callback be registered on this RPC so that an error could at least be logged?

src/modules/coral2_dws.py
```diff
@@ -401,58 +401,77 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
...
+    else:
+        LOGGER.debug("Marking rabbit %s as down, status is %s", resource_path, status)
+        payload = {"resource_path": resource_path, "status": "down"}
+        handle.rpc("sched-fluxion-resource.set_status", payload)
```
grondo, line 413: Same comment as above.

grondo (approved):
LGTM! Just a couple comments inline about dropping responses to RPCs.

src/modules/coral2_dws.py
jameshcorbett: Done!

jameshcorbett:
Comments addressed, setting MWP.

garlick:
Be vewey vewey quiet, we are dwaining wabbits!

vsoch:
:raised_hands: :raised_hands: :raised_hands:

PR #124, opened 2024-02-09
configure: downgrade required kubernetes version

jameshcorbett:
Problem: LC EAS3 clusters have the python kubernetes package version 11.0.0, not anything higher.

The code appears to work with version 11, so downgrade the required version to 11.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -67,9 +67,9 @@ AS_IF([test "x$enable_pylint" = "xyes"], [
 ])
 AM_CONDITIONAL([ENABLE_PYLINT], [test "x$PYLINT" = "xpylint"])
 AM_CHECK_PYMOD(kubernetes,
-                  [StrictVersion(kubernetes.__version__) >= StrictVersion('12.0.0')],
+                  [StrictVersion(kubernetes.__version__) >= StrictVersion('11.0.0')],
                   ,
-                  [AC_MSG_ERROR([could not find python module kubernetes, version 12.0.0+ required])]
+                  [AC_MSG_ERROR([could not find python module kubernetes, version 11.0.0+ required])]
                 )
 
 # Checks for header files.


grondo (approved):
LGTM!

wihobbs (approved):
LGTM, thanks!

FYI, Kubernetes has released a python client up to version 0.29.0, which corresponds to Kubernetes 1.29.0. They currently support 1.27-1.29 (and, I assume, the corresponding python clients, though I can't find any docs that explicitly say so). 

Anyway, my point is that we're lagging behind a bit on tioga -- ought we put in a ticket to LC to update the version?

jameshcorbett:
> ought we put in a ticket to LC to update the version?

Well, depending on the python install we targeted, the process would be different. The python relevant here is `/usr/bin/python` which I think means it comes from TOSS and I think getting Python packages upgraded there is hard? Relatively harder than the `/usr/bin/tce` installations. But either way I think it's difficult, and it seems to be working as-is so I'm not too worried. Any thoughts @grondo ?

grondo:
>  Any thoughts @grondo ?

Looks like the kubernetes package is coming from EPEL8 (Redhat Extra Packages for Enterprise Linux), so if you need the newer version we'd either have to ask Red Hat/Fedora to update the version in EPEL, or we'd need to start building a new version in the buildfarm. Either way you'd need to provide the reason that 11.0.0 doesn't work.

Just a note about OS distributions: When maintaining an OS distribution like TOSS, one needs to ensure a stable environment for the OS lifetime. Therefore, updates in OS packages are typically only done to fix bugs, security issues, or in rare cases to update to a newer version because a patched package does not work anymore or isn't supported, or there is only one user. This is why stable OSes are not in a "race to the latest version".

My point being that if the current python3-kubernetes package works, then it is probably best to continue to support that version. We don't want to update packages just because there is a new version. That way we don't update to the latest and greatest and then break any other users of that package. That being said, it sounds like versions of this package are somewhat compatible with eachother, and it seems like there are probably limited users
```console
$ rpm -q --whatrequires python3-kubernetes
python3-rosetta-dev-centos-2.1.0-3.noarch
```
so if you do need a newer version in the future, it might be acceptable to request it (or build the required version yourself in the toss buildfarm)

PR #127, opened 2024-02-12
shell: update cray_pals to edit LD_LIBRARY_PATH when under a source tree

### LINKED ISSUE #126: cray-pals fails to edit LD_LIBRARY_PATH when flux is started from a flux-core source tree
Problem: cray-pals fails to edit `pmi_library_path` from LD_LIBRARY_PATH when flux is run from a flux-core source tree.

This is because the simple PMI plugin fetches the directory using
```c
flux_conf_builtin_get ("pmi_library_path", FLUX_CONF_AUTO)
```
while cray-pals fetches it with:
```c
flux_conf_builtin_get ("pmi_library_path", FLUX_CONF_INSTALLED);
```

Since only the matching directory is removed by cray-pals, the removal only works when the FLUX_CONF_AUTO heuristic determines that the "installed" paths should be used.

To fix, change the cray-pals flag to FLUX_CONF_AUTO.

wihobbs:
Problem: cray-pals fails to edit pmi_library_path (from LD_LIBRARY_PATH) when flux is run from a flux-core source tree.

Change cray-pals to fetch the directory using FLUX_CORE_AUTO.

Fixes #126


wihobbs:
Yeah, I thought that title might be too long. Fixing...

jameshcorbett (approved):
LGTM, ok to set MWP?

wihobbs:
Thanks @jameshcorbett!

PR #130, opened 2024-03-01
test: request too much rabbit storage

jameshcorbett:
Fluxion should reject jobs that request an infeasible amount of rabbit storage. However, [this commit](https://github.com/jameshcorbett/flux-coral2/commit/297f1fc167e00f353016e26333a3d33a85c54e68) disabled Fluxion scheduling of rabbits. Revert that commit, fix a problem with reverting that commit, and add a test for the expected behavior.


vsoch:
There is too much wabbit storage in this test! :rabbit:  :rabbit: :carrot: :carrot: 

I will (again) see myself out.

cmoussa1 (approved):
LGTM!

PR #132, opened 2024-03-01
dws: enforce minimum rabbit allocation size

jameshcorbett:
This PR fixes one problem with rabbit allocations and alleviates another problem.

The fixed problem is that rabbit allocations have an implicit minimum size: allocations smaller than around 2 GiB fail or hang. Fix this by enforcing a configurable minimum size that defaults to 4GiB.

The alleviated problem is that lustre allocations are always split `N` ways for `N` the number of compute nodes. This means that sometimes a single rack will have many lustre allocations on it for a single job and lustre file system, when ideally all those allocations would be combined into one, thus saving some NVMe namespaces. However, this is a bit tricky due to the way DWS requires allocations to be specified; it's impossible to specify an allocation size for one rack independent of the size of the allocation on the other racks. However, allocation fragmentation can be reduced by finding the GCD of nodecounts on each rack and scaling up the allocation size accordingly.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -38,6 +38,7 @@ LOGGER = logging.getLogger(__name__)
 WORKFLOWS_IN_TC = set()  # tc for TransientCondition
 WORKFLOW_NAME_PREFIX = "fluxjob-"
 WORKFLOW_NAME_FORMAT = WORKFLOW_NAME_PREFIX + "{jobid}"
+_MIN_ALLOCATION_SIZE = 4  # minimum rabbit allocation size
 
 
 class WorkflowInfo:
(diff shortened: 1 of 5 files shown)


cmoussa1 (approved):
LGTM!

grondo:
One minor suggestion: PR titles end up being included as a commit message of sorts because of the merge commit. As such, the PR title should follow standard practice for commit message titles, in that there's an attempt to describe what is being changed in the imperative. I'm not sure if it is the case here, but in flux-core the merge commit messages also go in the release notes, so it is good if the PR title also makes a nice release note entry. 

My suggestion would be to expand the PR title a bit so readers know what "miminum allocation size" means and it is in the imperative sense, such as "Enforce a minumum allocation size on rabbits of 4GiB". 

This is just a suggestion and feel free to ignore me!

jameshcorbett:
Oh, thanks @grondo, I didn't know that about PR titles. Fixed and setting MWP.

PR #133, opened 2024-03-01
dws2jgf: rabbits as compute nodes

jameshcorbett:
El Capitan currently only has rabbit nodes and gateway nodes, no compute nodes. This makes testing the rabbits difficult. The best we can do for now is to treat rabbit nodes as compute nodes. This requires some changes to how the Fluxion resource graph is built for rabbits.

I think once El Capitan gets a couple dozen compute nodes, this won't be useful any more.

### DIFF
--- a/src/cmd/flux-dws2jgf.py
+++ b/src/cmd/flux-dws2jgf.py
@@ -10,6 +10,7 @@ import itertools
 import flux
 from flux.idset import IDset
 from flux.hostlist import Hostlist
+from flux.idset import IDset
 from fluxion.resourcegraph.V1 import (
     FluxionResourceGraphV1,
     FluxionResourcePoolV1,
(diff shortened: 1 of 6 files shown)


src/cmd/flux-dws2jgf.py
```diff
@@ -223,19 +224,10 @@ def get_node_children(r_lite):
...
-                for i in range(int(low), int(high) + 1):
-                    properties = rank_to_property.setdefault(i, [])
-                    properties.append(prop_name)
-            else:
-                properties = rank_to_property.setdefault(rank, [])
-                properties.append(prop_name)
+    for prop_name, idset_str in properties.items():
+        for rank in IDset(idset_str):
+            properties = rank_to_property.setdefault(rank, [])
+            properties.append(prop_name)
```
cmoussa1, line 230: nice cleanup!!

cmoussa1 (approved):
LGTM 👍

PR #135, opened 2024-03-04
dws2jgf: fix node attributes

jameshcorbett:
Problem: flux-dws2jgf writes out vertex properties as a python list,
but Fluxion's JGF reader expects a JSON object.

Write out properties as keys in a mapping with empty string values.

See also https://github.com/flux-framework/flux-sched/pull/1149.


cmoussa1 (approved):
LGTM 👍

PR #140, opened 2024-03-27
dws: add support for draining Offline nodes

### LINKED ISSUE #139: Draining compute nodes that lose connection with their rabbit
Sometimes compute nodes can lose their connection with their rabbits. When this happens they should be drained. Although compute-only jobs could still run on those nodes, Brian Behlendorf and @garlick seemed to agree in separate offline discussions that draining the nodes will get admins to investigate; if the nodes were not drained, they might not notice.

jameshcorbett:
Problem: compute nodes can lose their link with their local rabbit. If that happens, the node should be drained with a reasonable message so admins can investigate.

Kubernetes Storage resources list the status of the links between rabbits and their compute nodes. Watch the status and drain nodes if their status changes to 'Offline'.

Fixes #139 .

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -443,6 +443,25 @@ def _workflow_state_change_cb_inner(workflow, jobid, winfo, handle, k8s_api):
         WORKFLOWS_IN_TC.discard(winfo)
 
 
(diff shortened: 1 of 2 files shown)


cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #141, opened 2024-03-27
dws: flux finalizer

### LINKED ISSUE #138: dws: add a "flux" finalizer to Kubernetes workflow resources
When a Workflow is deleted manually, the `coral2_dws` service loses its trigger to release the `dws-epilog` jobtap epilog action, so jobs are stuck in CLEAN while still holding resources. Blake Devcich suggested:

> While users shouldn't be doing much of this at all, I think you should consider putting a finalizer on workflows. 
> Once flux has done all its cleanup, (but before the delete) it could remove that finalizer. That way no one can pull the rug out from underneath flux unless they also delete that finalizer.

> Finalizers are part of the the metadata section and it's a list of strings. In order for a k8s object to be deleted, there must not be any finalizers.
> 
> ```
> $ kubectl get workflow fluxjob-57852485238259712 -oyaml | yq '.metadata.finalizers'
> - dataworkflowservices.github.io/workflow
> - nnf.cray.hpe.com/nnf_workflow
> ```
> 
> We put those finalizers on workflows until it's time to clean up and then remove them so the object can delete.

jameshcorbett:
When a Workflow is deleted manually, the coral2_dws service loses its trigger to release the dws-epilog jobtap epilog action, so jobs are stuck in CLEAN while still holding resources. Adding a finalizer prevents the workflow from being deleted.

Fixes #138.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -39,6 +39,7 @@ WORKFLOWS_IN_TC = set()  # tc for TransientCondition
 WORKFLOW_NAME_PREFIX = "fluxjob-"
 WORKFLOW_NAME_FORMAT = WORKFLOW_NAME_PREFIX + "{jobid}"
 _MIN_ALLOCATION_SIZE = 4  # minimum rabbit allocation size
+_FINALIZER = "flux-framework.readthedocs.io/workflow"
 
 
 class WorkflowInfo:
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM! 👍 You might consider adding a brief description to the PR as well.

PR #142, opened 2024-03-30
dws-jobtap: fix race conditions with exception-raising

jameshcorbett:
Problem: the `dws-jobtap` plugin has a race condition in its exception-raising logic.


grondo (approved):
LGTM!

jameshcorbett:
Thanks! Set MWP.

PR #143, opened 2024-04-03
dws: fix floating-point truncation

jameshcorbett:
Problem: when splitting an allocation across multiple rabbits, the size of each individual allocation must not be rounded down or the total size may not meet the minimum required.

Round up instead of rounding down when splitting allocations.


cmoussa1 (approved):
LGTM! 👍

jameshcorbett:
Thanks! Setting MWP.

PR #150, opened 2024-04-17
Handle rabbits with no computes

jameshcorbett:
Problem: flux-dws2jgf assumes storages always have a
`.status.access.computes` field, but recent experience on rzvernal
has shown this to be false.

Make .status.access.computes optional.


cmoussa1 (approved):
LGTM! IMHO, you could probably squash these two commits into just one since they both look to be doing the same thing. 👍 Feel free to ignore, though

jameshcorbett:
Makes sense, done and setting MWP!

PR #151, opened 2024-04-18
dws: fix rabbit name manipulation logic

jameshcorbett:
Problem: coral2_dws tries to strip a 'rabbit-' prefix from the names
of rabbit vertices it finds in JGF. However, it uses the str.strip()
method incorrectly and in fact strips the characters 'r', 'a', 'b',
'i', 't', and '-' from the beginning and end of the rabbit name.

Strip the prefix correctly.


cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks! Set MWP.

PR #152, opened 2024-04-18
k8s watch exception handling

jameshcorbett:
Problem: while watching for k8s resources, a resourceversion > 0
is required. When there are only old resources lying around and
no new ones, the resourceversion will be > 0 but too old for
kubernetes to accept it, and an exception may be raised.

Catch the exception and set the resource version to 0.

### DIFF
--- a/src/python/flux_k8s/watch.py
+++ b/src/python/flux_k8s/watch.py
@@ -3,6 +3,8 @@ import logging
 
 from flux.constants import FLUX_MSGTYPE_REQUEST
 import kubernetes as k8s
+from kubernetes.client.exceptions import ApiException
+
 
 from flux.core.watchers import fd_handler_wrapper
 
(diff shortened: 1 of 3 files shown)


src/python/flux_k8s/watch.py
```diff
@@ -20,16 +22,32 @@ def __init__(self, api, crd, resource_version, cb, *args, **kwargs):
         self.cb_kwargs = kwargs
 
     def watch(self):
+        """Watch resource, firing off callbacks.
+
+        When there are only old resources lying around and no new ones,
+        the resourceversion will be > 0 but too old for kubernetes to accept it.
+
+        Some versions of kubernetes (newer ones) throw an exception, others
+        just return an ERROR event. Handle both cases. By resetting the
```
cmoussa1, line 31: I think this is a sentence fragment here, i.e no need for a period in between "cases" and "by"

cmoussa1 (approved):
LGTM! One minor suggestion about fixing one of the multiline comments. 👍

src/python/flux_k8s/watch.py
jameshcorbett: Good catch thanks!

jameshcorbett:
Comment addressed, rebased with master, setting MWP.

PR #153, opened 2024-04-27
dws2jgf: use sorted hostlist to determine ranks

jameshcorbett:
Problem: the dws2jgf script does not sort the hostlist it uses to determine the ranks of nodes, leading to mixups if the hostlist is not already sorted.

Sort the hostlist.

### DIFF
--- a/src/cmd/flux-dws2jgf.py
+++ b/src/cmd/flux-dws2jgf.py
@@ -299,7 +299,7 @@ def main():
 
     input_r = json.load(sys.stdin)
     nnfs = [x for x in get_storage()["items"]]
-    r_hostlist = Hostlist(input_r["execution"]["nodelist"])
+    r_hostlist = Hostlist(input_r["execution"]["nodelist"]).sort()
     dws_computes = set(
         compute["name"]
         for nnf in nnfs


src/cmd/flux-dws2jgf.py
```diff
@@ -299,7 +299,7 @@ def main():
 
     input_r = json.load(sys.stdin)
     nnfs = [x for x in get_storage()["items"]]
-    r_hostlist = Hostlist(input_r["execution"]["nodelist"])
+    r_hostlist = Hostlist(input_r["execution"]["nodelist"]).sort()
```
grondo, line 302: The hostlist in Rv1 should be in rank order (See [RFC 20](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html#nodelist)), so should not be sorted if you are using it to map hostnames to ranks.

Some caveats:
- `hostlist[0]` corresponds to the lowest numbered rank entry in `R_lite`, not necessarily rank 0
- There's a bug in flux-core up to v0.61.2 where the encoding Rv1 is incorrectly sorted (ranks are sorted and the hostlist is not), so that may be why you ran into an issue? Not sure what to do about that -- it is an unfortunate bug.

jameshcorbett: Oh, yeah OK I guess I ran into that bug on tioga:
```
[corbett8@tioga10:~]$ cat /etc/flux/system/R | jq .execution.nodelist
[
  "tioga[6,10-11,18-41,14-17,42-43]"
]
[corbett8@tioga10:~]$ ssh tioga20 flux getattr rank
9
[corbett8@tioga10:~]$ ssh tioga15 flux getattr rank
4
```

jameshcorbett: To your first point, when would the R for the system instance not contain all ranks?

jameshcorbett:
Given @grondo 's comments, these changes are not necessary. Closing this PR.

src/cmd/flux-dws2jgf.py
grondo: Oh, I can't think of when the `resource.R` wouldn't have all ranks, but was thinking of all other situations -- notably the scheduler may not receive an R with all ranks from `resource.acquire` if some are excluded.

PR #154, opened 2024-05-08
docker: mount $HOME as $HOME

jameshcorbett:
Problem: flux-core and -sched both have docker-run-checks scripts that mount the $HOME directory as $HOME, not as /home/$USER (as this project does).

Change to be consistent with those projects.


grondo (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #155, opened 2024-05-10
Revert docker home mount and change hostname

jameshcorbett:
PR #154 was opened too hastily. By default, `kubectl` (which is used all over the place in flux-coral2 tests) looks for a kubeconfig file in `$HOME/.kube/config`. And the flux-coral2 testsuite has logic in place, such as [here](https://github.com/flux-framework/flux-coral2/blob/master/t/sharness.d/03-kube.sh), to manipulate kubernetes into working properly which was affected by PR #154. It caused test failures which were only visible when running with kubernetes (which CI doesn't do).

The changes introduced by #154 are still desirable (see issue #156), but they will need to be introduced with more extensive changes which hopefully would make the kubernetes manipulations more straightforward.

### DIFF
--- a/src/test/docker/docker-run-checks.sh
+++ b/src/test/docker/docker-run-checks.sh
@@ -12,7 +12,7 @@ BASE_DOCKER_REPO=fluxrm/flux-sched
 
 IMAGE=focal
 JOBS=2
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM

PR #157, opened 2024-05-10
Resource graph: rack-parent ssds

jameshcorbett:
As noted [here](https://github.com/flux-framework/flux-sched/discussions/1189), the existing Fluxion graph layout imposes a serious limitation on scheduling: every node is forced to be on a different rack.

By changing the resource graph to remove intermediate "rabbit" vertices between "rack" and "ssd", this limitation can be worked around. It also opens up possibilities for better Fluxion scheduling of lustre-only rabbit jobs.


cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #163, opened 2024-05-14
dws: change exception import path

jameshcorbett:
Problem: some older versions of kubernetes do not have a 'kubernetes.client.exceptions' module which 'watch.py' attempts to import from.

Add a try/except around the import, and if the import fails, create a dummy exception with the same name.

### DIFF
--- a/src/python/flux_k8s/watch.py
+++ b/src/python/flux_k8s/watch.py
@@ -5,8 +5,6 @@ from flux.constants import FLUX_MSGTYPE_REQUEST
 import kubernetes as k8s
 from kubernetes.client.rest import ApiException
 
-from flux.core.watchers import fd_handler_wrapper
-
 
 LOGGER = logging.getLogger(__name__)
 
--- a/src/python/flux_k8s/watch.py
+++ b/src/python/flux_k8s/watch.py
@@ -3,8 +3,7 @@ import logging
 
 from flux.constants import FLUX_MSGTYPE_REQUEST
 import kubernetes as k8s
-from kubernetes.client.exceptions import ApiException
-
+from kubernetes.client.rest import ApiException
 
 from flux.core.watchers import fd_handler_wrapper
 


grondo (approved):
LGTM!

vsoch:
What is the version, and is the action of the exception class meaningful? The oldest I see ([this release](https://github.com/kubernetes-client/python/blob/adf0faa91036aad0790746ef8c44114f4f50c56b/kubernetes/client/rest.py#L298)) it looks like the import path is just different - from `kubernetes.client.rest`.

vsoch:
And a question I have is if your `APIException` type is going to be considered the same as [the one that watch would raise](https://github.com/kubernetes-client/python/blob/94e42113a1fe5c580917decacdde879eab7406b3/kubernetes/base/watch/watch.py#L195).  For example (this is probably a dumb example, but similar in concept):

```python
# Here is function with a known exception type, ValueError
def do_bad_things():
    raise ValueError("You did bad things.")

# And here I am making a class on the fly
class ValueError(Exception):
    pass

The function is tied to the one it was defined with
do_bad_things()

ValueError: You did bad things.
```
```console
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 do_bad_things()

Input In [1], in do_bad_things()
      1 def do_bad_things():
----> 2     raise ValueError("You did bad things.")
```
So maybe (if possible) a sanity check to see if when there is actually an API error it is going to hit that case. If not and you don't want to import the other path, you can catch a general `Exception` and then parse it to determine the name of the class is `APIException`.

jameshcorbett:
Great catch @vsoch it looks like the import path you mentioned, `kubernetes.client.rest`, works for versions v30.1.0 down to v11.0.0 which I think is pretty solid, so I'm going to use that path. At some point in there they changed where the exception is defined, but it can still be imported from the original location.

I also tacked on a housekeeping commit.

jameshcorbett:
Ok, fixed up the commit title and first commit message. @grondo @vsoch can one of you take a look again?

grondo (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #164, opened 2024-05-15
dws2jgf: change cluster-name logic

jameshcorbett:
Problem: at the moment admins should supply the name of the cluster through a command-line argument when running dws2jgf. But they will probably run it from the management node of a cluster, and most LC clusters have nodes with names like 'nameXXXXX' for XXXXX some decimal number.

Simplify things for them by making the cluster name in the resource graph default to the hostname stripped of numerals.

### DIFF
--- a/src/cmd/flux-dws2jgf.py
+++ b/src/cmd/flux-dws2jgf.py
@@ -6,6 +6,7 @@ import json
 import re
 import logging
 import itertools
+import socket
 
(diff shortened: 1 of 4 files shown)


cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP.

PR #168, opened 2024-06-29
dws: enforce k8s finalizer removal

jameshcorbett:
Problem: users have reported on elcap that some of their workflows hang in deletion because flux attempts to delete it without first removing the finalizer.

Add additional logic for removing Flux's finalizer from workflows.

Fixes #165.

PR #169, opened 2024-06-29
dws: fix Postrun workflow error hang

jameshcorbett:
Problem: if a workflow goes to status: Error, coral2_dws will raise
an exception. For most states the exception will transition the job
to CLEANUP, which eventually moves the workflow to Teardown.
However, if the job is already in CLEANUP and is being held in the
dws-epilog epilog when the exception is thrown, the exception has
no effect and the epilog will never be finished because the workflow
does not progress.

If a workflow moves to status: Error in PostRun or DataOut states,
move it directly to Teardown.


cmoussa1 (approved):
LGTM 👍

jameshcorbett:
Rebased after #168 was merged and dropped the draft label, no code changes. Thanks for the review! Setting MWP.

PR #177, opened 2024-07-11
Exception log improvements

jameshcorbett:
Miscellaneous log and exception improvements to satisfy my HPE overlords

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -345,7 +345,7 @@ def workflow_state_change_cb(event, handle, k8s_api, disable_fluxion):
         jobid = int(flux.job.JobID(workflow["spec"]["jobID"]))
         workflow_name = workflow["metadata"]["name"]
     except KeyError:
-        LOGGER.exception("Invalid workflow in event stream: ")
+        LOGGER.exception("Invalid event %s in workflow stream: ", event)
         return
     if not workflow_name.startswith(WORKFLOW_NAME_PREFIX):
         LOGGER.warning("unrecognized workflow '%s' in event stream", workflow_name)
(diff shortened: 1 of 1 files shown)


jameshcorbett:
@bdevcich I can't tag you as a reviewer for some reason however it would be most appreciated

cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks! Setting MWP

src/modules/coral2_dws.py
```diff
@@ -145,6 +143,17 @@ def remove_finalizer(workflow_name, k8s_api, workflow=None):
         )
 
 
+def move_workflow_to_teardown(winfo, k8s_api, workflow=None):
+    """Helper function for moving a workflow to Teardown."""
+    if workflow is None:
+        workflow = k8s_api.get_namespaced_custom_object(*WORKFLOW_CRD, winfo.name)
+    LOGGER.debug("Moving workflow %s to Teardown, dump is %s", winfo.name, workflow)
```
bdevcich, line 150: what log does this end up in? event log?

bdevcich: Dang. Is there any way to get this stuff into something like the event log or something accessible from flux?

jameshcorbett: Let me see. My first take is that the eventlog would be bad because the workflow is so big, it would overwhelm all the other entries and make it look really cluttered. But maybe there is something else we could do.

bdevcich: Would it be possible to add a new entry for rabbit info similar to event log?

jameshcorbett: Feel free to comment on #180

PR #178, opened 2024-07-13
dws: dump datamovements to logs

jameshcorbett:
Problem: my HPE overlords want the coral2-dws service to log `nnfdatamovements`.


cmoussa1 (approved):
LGTM!

PR #179, opened 2024-07-13
dws: add marker for prolog to enable the nnf-dm daemon

### LINKED ISSUE #167: Start nnf-dm daemon on compute nodes when job requests it
Similar to #166, there is a daemon that we should start and stop at different points in the job lifecycle. However, the daemon mentioned in that issue (called `nnf-clientmount`) should be stopped when a job begins execution and started after it finishes execution. This daemon, `nnf-dm`, will need to be started for some rabbit jobs (but not all) when the job starts and then stopped when the job finishes. There is information in kubernetes `directivebreakdown` resources indicating whether a job will need this daemon active.

Currently the daemon runs all the time. A simple improvement would be to leverage the solution in #166 and put logic in the prolog to only start the daemon for rabbit jobs, and have the epilog always stop the daemon. However, there will still be plenty of rabbit jobs that do not need the daemon running.

Perhaps some additional context can be added to the `dws_environment` eventlog entry to indicate whether a rabbit job needs the daemon or not, or to the `rabbits` memo, or similar, and the prolog could check for it, since it already waits for that event (as described in #166).

jameshcorbett:
Problem: There is no marker that a job needs the nnf data movement
daemon to be running during its duration. Without a marker, an
administrative prolog cannot tell whether to start the daemon.

Add a boolean to the dws-environment event that indicates whether
the nnf-dm daemon is needed. The administrative prolog can watch for
this event and its context.

Fixes #167.

@grondo, what would be the best way for a shell script (the administrative prolog) to watch for the boolean? It is already watching for the `dws_environment` event, but now it also needs to know whether the `copy-offload` entry in the context is true or false.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -576,14 +576,16 @@ static void resource_update_msg_cb (flux_t *h,
     json_int_t jobid;
     json_t *resources = NULL;
     json_t *jobspec;
+    int copy_offload;
 
-    if (flux_msg_unpack (msg, "{s:I, s:o}", "id", &jobid, "resources", &resources)
+    if (flux_msg_unpack (msg, "{s:I, s:o, s:b}", "id", &jobid, "resources", &resources, "copy-offload", &copy_offload)
         < 0) {
         flux_log_error (h, "received malformed dws.resource-update RPC");
         return;
     }
     if (!(jobspec = flux_jobtap_job_aux_get (p, jobid, "dws_jobspec"))
-        || json_object_set (jobspec, "resources", resources) < 0) {
+        || json_object_set (jobspec, "resources", resources) < 0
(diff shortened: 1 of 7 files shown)


grondo:
> what would be the best way for a shell script (the administrative prolog) to watch for the boolean? 

How does the prolog wait for the `dws_environment` event now? 

One way would be to dump the eventlog as json and process with `jq`:
```
flux job eventlog -f json ƒ9zRvw7KVq | jq 'select(.name == "dws_environment")| .context["copy-offload"]'
```
This would print `true` or `false` I think if the `dws_environment` event is found in the eventlog (and `copy-offload` exists in the context), otherwise it would fail.

You could also do this in two stages by saving the eventlog or the `dws_environment` entry to a file, then process with jq as a separate step.

There may other ways as well (e.g. `flux job wait-event` has a `--match-context` option.

jameshcorbett:
Perfect, thank you!

cmoussa1 (approved):
LGTM 👍👍

jameshcorbett:
Thanks! I just force-pushed to add an extra test to `t1002-dws-workflow-obj.t`. Setting MWP.

Issue #171, opened 2024-07-02
Allow more MDT flexibility in Rabbit lustre allocations


jameshcorbett:
Problem: by default, Flux creates one MDT per rabbit for lustre file systems. However, `directivebreakdown` resources list some info which may indicate that fewer (or more) MDTs should be created.

@behlendorf said:

> We really haven't done much testing with multiple rabbits but being able to control the number of MDTs and OSTs is going to be important at scale.

Further, creating one MDT per rabbit

> [Is] going to be an issue at scale.  Lustre has issues beyond 50'ish MDTs in a filesystems, it should work but performance will get worse as MDTs are added.

Flux should look at directivebreakdowns to see if they offer hints on how many MDTs to create.

jameshcorbett:
Some documentation is here https://nearnodeflash.github.io/dev/guides/directive-breakdown/readme/

PR #181, opened 2024-07-15
dws: crudely enforce mdt count constraints

jameshcorbett:
Problem: as described in #171, creating many MDTs is a bad
for performance, and usually goes against what is explicitly
required by directivebreakdown resources. However, there is not
yet a good way to get Fluxion to handle MDT allocation.

Bypass Fluxion allocation completely, and tell DWS to create
exactly the number of allocations requested in the
.constraints.count field (which is usually found on MDTs).

Place the allocations on the rabbits which have the most compute
nodes allocated to the job.

This is intended to be only a temporary solution, since it
adds a new potential problem, in that some rabbit storage is
used which is not tracked by Fluxion. This could lead to
overallocation of resources, causing jobs to fail with errors.
However, this seems unlikely to occur in practice, since MDTs
are small and Fluxion always gives jobs more storage than they
asked for, so there should usually be some spare storage.

### DIFF
--- a/src/python/flux_k8s/directivebreakdown.py
+++ b/src/python/flux_k8s/directivebreakdown.py
@@ -1,3 +1,5 @@
+"""Module defining functions related to DirectiveBreakdown resources."""
+
 import enum
 import copy
 import functools
@@ -7,6 +9,8 @@ from flux_k8s.crd import DIRECTIVEBREAKDOWN_CRD
 
 
 class AllocationStrategy(enum.Enum):
+    """Enum defining different AllocationStrategies."""
+
     PER_COMPUTE = "AllocatePerCompute"
     SINGLE_SERVER = "AllocateSingleServer"
     ACROSS_SERVERS = "AllocateAcrossServers"
@@ -17,6 +21,7 @@ LUSTRE_TYPES = ("ost", "mdt", "mgt", "mgtmdt")
 
 
 def build_allocation_sets(breakdown_alloc_sets, nodes_per_nnf, hlist, min_alloc_size):
+    """Build the allocationSet for a Server based on its DirectiveBreakdown."""
     allocation_sets = []
     for alloc_set in breakdown_alloc_sets:
         storage_field = []
@@ -61,7 +66,6 @@ def apply_breakdowns(k8s_api, workflow, old_resources, min_size):
     """Apply all of the directive breakdown information to a jobspec's `resources`."""
     resources = copy.deepcopy(old_resources)
     breakdown_list = list(fetch_breakdowns(k8s_api, workflow))
-    per_compute_total = 0  # total bytes of per-compute storage
     if not resources:
         raise ValueError("jobspec resources empty")
     if len(resources) > 1 or resources[0]["type"] != "node":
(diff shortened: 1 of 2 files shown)


src/python/flux_k8s/directivebreakdown.py
```diff
@@ -41,19 +42,45 @@ def build_allocation_sets(breakdown_alloc_sets, nodes_per_nnf, hlist, min_alloc_
...
-                    {
-                        "allocationCount": nodes_per_nnf[rabbit_name] / nodecount_gcd,
-                        "name": rabbit_name,
-                    }
+            if "count" in alloc_set.get("constraints", {}):
+                # a specific number of allocations is required (generally for MDTs)
+                count = alloc_set["constraints"]["count"]
+                server_alloc_set["allocationSize"] = math.ceil(
+                    alloc_set["minimumCapacity"] / count
+                )
```
cmoussa1, line 50: probably just a dumb question, but could `count` ever be 0 in case the key-value lookup doesn't succeed for whatever reason? Could maybe be helpful to check before attempting this division, although Python generally doesn't freak out when attempting division by 0, just raises an error :)  I'll leave it up to you whether that could be helpful

src/python/flux_k8s/directivebreakdown.py
```diff
@@ -41,19 +42,45 @@ def build_allocation_sets(breakdown_alloc_sets, nodes_per_nnf, hlist, min_alloc_
...
+                    ):
+                        storage_field.append(
+                            {
+                                "allocationCount": 1,
+                                "name": name,
+                            }
+                        )
+                        count -= 1
+                        if count == 0:
+                            break
```
cmoussa1, line 67: if you change the `while` condition to `>= 1`, maybe you don't need to add this extra `if` check at the end of every loop?

cmoussa1 (approved):
LGTM! Just a couple of minor clarification questions. Nice work!

src/python/flux_k8s/directivebreakdown.py
jameshcorbett: I think `count` could be 0, but then something in the k8s environment must have been misconfigured and would need tweaking by an admin. I think the zerodivisionerror would be OK here. Good catch though.

src/python/flux_k8s/directivebreakdown.py
jameshcorbett: I don't see how `>=1` would be different than `> 0`? Also the `if` check is just to break out of the inner loop, which would otherwise proceed without regard to the value of `count`--I wanted to make sure that exactly `count` allocations were made.

cmoussa1: Gotcha - my mistake!! Thanks for the clarification

src/python/flux_k8s/directivebreakdown.py
cmoussa1: sounds great - just wanted to make sure. thanks!!

jameshcorbett:
Thanks for the review @cmoussa1 ! Setting MWP.

PR #182, opened 2024-07-16
dws: change rabbit memo to hostlist

jameshcorbett:
Problem: the flux-coral2 script adds a memo to rabbit jobs with a JSON list of the rabbits used. This is inconvenient and can potentially become large.

Use a hostlist instead.


cmoussa1 (approved):
👍 👍

jameshcorbett:
Thanks, setting MWP!

PR #187, opened 2024-07-25
Kvs workflow timings

jameshcorbett:
Two changes:

1) Save workflows to the KVS after the completion of every state except Teardown. So now you don't need to wait for the job to complete to look at your workflow
2) Save the `.status.elapsedTimeLastState` field of workflows to the KVS after the completion of every state except Teardown. The timing goes under the `rabbit_{state}_timing` key so to get the timing for DataIn, you could run `flux job info $JOBID rabbit_DataIn_timing`.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -147,18 +147,23 @@ def remove_finalizer(workflow_name, k8s_api, workflow):
         )
 
 
-def move_workflow_to_teardown(handle, winfo, k8s_api, workflow=None):
-    """Helper function for moving a workflow to Teardown."""
-    if workflow is None:
-        workflow = k8s_api.get_namespaced_custom_object(*WORKFLOW_CRD, winfo.name)
+def save_workflow_to_kvs(handle, jobid, workflow):
+    """Save a workflow to a job's KVS, ignoring errors."""
     try:
-        kvsdir = flux.job.job_kvs(handle, winfo.jobid)
+        kvsdir = flux.job.job_kvs(handle, jobid)
         kvsdir["rabbit_workflow"] = workflow
         kvsdir.commit()
     except Exception:
         LOGGER.exception(
-            "Failed to update KVS for job %s: workflow is", winfo.jobid, workflow
+            "Failed to update KVS for job %s: workflow is", jobid, workflow
         )
+
+
+def move_workflow_to_teardown(handle, winfo, k8s_api, workflow=None):
+    """Helper function for moving a workflow to Teardown."""
+    if workflow is None:
+        workflow = k8s_api.get_namespaced_custom_object(*WORKFLOW_CRD, winfo.name)
+    save_workflow_to_kvs(handle, winfo.jobid, workflow)
     if LOGGER.isEnabledFor(logging.INFO):
         try:
             api_response = k8s_api.list_cluster_custom_object(
@@ -446,6 +451,7 @@ def _workflow_state_change_cb_inner(
                 "copy-offload": copy_offload,
             },
         ).then(log_rpc_response)
+        save_workflow_to_kvs(handle, jobid, workflow)
     elif state_complete(workflow, "Setup"):
         # move workflow to next stage, DataIn
         move_workflow_desiredstate(winfo.name, "DataIn", k8s_api)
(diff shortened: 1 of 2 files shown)


src/modules/coral2_dws.py
```diff
@@ -446,12 +459,15 @@ def _workflow_state_change_cb_inner(
                 "copy-offload": copy_offload,
             },
         ).then(log_rpc_response)
+        save_workflow_to_kvs(handle, jobid, workflow)
     elif state_complete(workflow, "Setup"):
         # move workflow to next stage, DataIn
         move_workflow_desiredstate(winfo.name, "DataIn", k8s_api)
+        save_workflow_to_kvs(handle, jobid, workflow)
```
bdevcich, line 466: I think you'll want to do this before the `move_workflow_desiredstate`

t/t1002-dws-workflow-obj.t
```diff
@@ -150,7 +150,13 @@ test_expect_success 'job submission with valid DW string works' '
 	flux job info ${jobid} rabbit_workflow | \
 		jq -e ".metadata.name == \"fluxjob-$(flux job id ${jobid})\"" &&
 	flux job info ${jobid} rabbit_workflow | jq -e ".spec.wlmID == \"flux\"" &&
-	flux job info ${jobid} rabbit_workflow | jq -e ".kind == \"Workflow\""
+	flux job info ${jobid} rabbit_workflow | jq -e ".kind == \"Workflow\"" &&
+	flux job info ${jobid} rabbit_Proposal_timing &&
```
bdevcich, line 154: I realize the stage here matches the capitalization that we use for workflows, but should they be lowercase to be consistent with the rest of the flux UI?

src/modules/coral2_dws.py
jameshcorbett: The `workflow` that's passed in to `save_workflow_to_kvs` is a Python dictionary that isn't changed by the call to `move_workfow_desiredstate` (if that's what you're concerned about) so it should be OK I think.

t/t1002-dws-workflow-obj.t
jameshcorbett: I don't really have a strong opinion, for the one-word states like Proposal I think you're right, but I was thinking that `DataOut` is easier to read than `dataout`.

jameshcorbett: Maybe I will just lowercase it though... Anyone else have thoughts? @cmoussa1 ?

garlick:
Remember all versions of a KVS key are retained and contribute to growth of the sqlite database.  Each version is a new blob object with a new hash - new blob gets written and metadata is updated with new blobref.  I'm not sure how big the data is or how many times it will get rewritten with this change, but just keep that in mind so it doesn't get out of hand.

Also, we do have an append operation in the KVS and [RFC 18](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html) which defines the eventlog format that's sometimes handy to use with KVS appends.  When appending data, each append is an independent blob, and hashes are appended to the KVS metadata (all versions of which are kept but it is slower growth).

Anyway, just to keep in mind, not to impede your forward progress.

jameshcorbett:
@garlick thanks for the reminder, I had forgotten. This PR overwrites a ~4Kb value like six times? I could definitely cut that down to one overwrite.

garlick:
:shrug: up to you but maybe a good idea if it's not buying you much.

src/modules/coral2_dws.py
bdevcich: Yes, that was my concern. Thanks for clarifying.

bdevcich (approved):

t/t1002-dws-workflow-obj.t
cmoussa1: I think I'm of the opinion that all lowercase is okay to go with, and would be consistent with the rest of the UI. If it's possible (and not too much of a hassle), you could always separate with underscores `_`, but either approach is fine with me.

cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks for the input everyone. Force-pushed to cut the number of KVS overwrites down to 1. Now the workflow is saved only after Proposal and DataOut. Setting MWP.

PR #189, opened 2024-08-03
DWS: handling for Lustre MGTs

### LINKED ISSUE #186: Rabbit DirectiveBreakdown AllocateSingleServer not working
Trying to create a standalone MGS through flux doesn't work. The "servers" resource is filled in with an empty "storage" field.

A standalone MGS can be created with "standaloneMgtPoolName" set to the pool name in the NnfStorageProfile. When a "#dw create_persistent type=lustre ..." workflow is created, a DirectiveBreakdown is made with a single allocation set asking for space for the MGT.

NnfStorageProfile:
```
...
data:
  default: false
  lustreStorage:
    capacityMgt: 5GiB
    combinedMgtMdt: false
    exclusiveMdt: false
    standaloneMgtPoolName: test-pool
...
```

DirectiveBreakdown:
```
...
status:
  ready: true
  storage:
    allocationSets:
    - allocationStrategy: AllocateSingleServer
      constraints:
        colocation:
        - key: lustre-mgt
          type: exclusive
        count: 1
        labels:
        - dataworkflowservices.github.io/storage=Rabbit
      label: mgt
      minimumCapacity: 5368709120
```

Flux fills in the "Servers" resource as follows:
```
    "spec": {
        "allocationSets": [
            {
                "allocationSize": 5368709120,
                "label": "mgt",
                "storage": []
            }
        ]
    },
```

From looking at the DWS code, it looks like the AllocateSingleServer case is not handled here: https://github.com/flux-framework/flux-coral2/blob/master/src/python/flux_k8s/directivebreakdown.py#L34 resulting in the empty "Storage" field.

I haven't tested it, but this would probably also cause a problem for a Lustre file system that allocated an MGT as part of the job (i.e., ExternalMgs not set).

jameshcorbett:
Problem: Flux does not handle workflows that require Lustre MGTs,
or more specifically any DirectiveBreakdown with an allocation type
of 'AllocateSingleServer'. However, Flux handles MDTs and other
allocations which have a 'count' constraint of 1, which is the same
thing effectively.

Handle 'AllocateSingleServer' allocations the same way that 'count'
constraints are handled (see #181 ), which is to pick a single rabbit and
place the allocation there, ignoring Fluxion's allocation choices
somewhat.

Fixes #186 .

### DIFF
--- a/t/t1002-dws-workflow-obj.t
+++ b/t/t1002-dws-workflow-obj.t
@@ -371,6 +371,22 @@ test_expect_success 'job submission with persistent DW string works' '
 	flux job wait-event -vt 30 ${jobid} clean
 '
 
+test_expect_success 'job submission with standalone MGT persistent DW string works' '
+	kubectl get nnfstorageprofiles -nnnf-system default -ojson | \
+		jq ".data.lustreStorage.standaloneMgtPoolName = \"mypool\" |
+		.metadata.name = \"mypoolprofile\" | .data.default = false |
+		.data.lustreStorage.combinedMgtMdt = false" \
+		| kubectl apply -f - &&
+	flux run --setattr=system.dw="#DW create_persistent type=lustre name=project1 profile=mypoolprofile" \
+		-N1 -n1 -c1 hostname &&
+	jobid=$(flux submit --setattr=system.dw="#DW destroy_persistent name=project1" \
+		-N1 -n1 -c1 hostname) &&
+	flux job wait-event -vt 30 -m description=${EPILOG_NAME} \
+		${jobid} epilog-start &&
+	flux job wait-event -vt 30 ${jobid} clean &&
+	kubectl delete nnfstorageprofiles -nnnf-system mypoolprofile
+'
+
 test_expect_success 'dws service script handles restarts while a job is running' '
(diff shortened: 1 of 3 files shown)


t/t1002-dws-workflow-obj.t
```diff
@@ -369,6 +369,22 @@ test_expect_success 'job submission with persistent DW string works' '
 	flux job wait-event -vt 30 ${jobid} clean
 '
 
+test_expect_success 'job submission with standalone MGT persistent DW string works' '
```
cmoussa1, line 372: Ignore this if I'm misunderstanding the acronyms - "MDT" here instead of "MGT"?

cmoussa1 (approved):
LGTM! One minor spelling comment.

t/t1002-dws-workflow-obj.t
cmoussa1: Whoops, yeah you can probably just ignore this. Just realized "MGT" is in the commit title - my bad!!

jameshcorbett: No you're right, it looks like one commit has MDT and one has MGT. They should both be MGT. Good catch! I'll fix and set MWP.

PR #191, opened 2024-08-05
dws: save workflow teardown timings to KVS

jameshcorbett:
Problem: starting in #187, the elapsed time for every state except Teardown is saved to a job's KVS. There is no reason for this exclusion, and the teardown timing may be useful or interesting.

Save the timing to the KVS and add a brief check for the entry in tests.


jameshcorbett:
Thanks! Setting MWP.

PR #194, opened 2024-08-07
dws: increase timeout of shell plugin

### LINKED ISSUE #192: Failure to read eventlog in `dws_environment` shell plugin
Marty McFadden found that the `dws_environment` shell plugin failed repeatedly in his tests with the following error:

```
flux-shell[0]: ERROR: dws_environment: No 'dws_environment' event posted
flux-shell[0]: ERROR: dws_environment: Error reading DW environment from eventlog
flux-shell[0]: ERROR: plugin 'dws_environment': shell.init failed
flux-shell[0]: FATAL: shell_init: Invalid argument
```

The problem is intermittent, about a quarter of Marty's 600 jobs hit it. So I'm guessing it is hitting [the timeout](https://github.com/flux-framework/flux-coral2/blob/3a388c1bcdaf25bcd06b43f86c4bb32c1c2cdd23/src/shell/plugins/dws_environment.c#L96) waiting to read another event. I wonder if given the size of elcap, a higher timeout is necessary? The event is definitely there, it's posted before the dws jobtap prolog is removed, and also the administrative prolog waits for the same event to be posted. So I don't believe there is any race condition.

@grondo , any thoughts?

jameshcorbett:
Problem: as described in issue #192, on large, busy systems with lots of jobs, the dws_environment shell plugin's two-second timeout to fetch the first eventlog entry is too short and occasionally times out.

In Mark Grondona's tests to reproduce the issue, all succeeded within five seconds. However, to be safe, increase the timeout to thirty seconds.

Fixes #192.

### DIFF
--- a/src/shell/plugins/dws_environment.c
+++ b/src/shell/plugins/dws_environment.c
@@ -93,7 +93,7 @@ static int read_future (flux_shell_t *shell, flux_future_t *fut)
     json_t *env, *rabbit_mapping;
     const char *name, *event = NULL;
 
-    while (flux_future_wait_for (fut, 2.0) == 0
+    while (flux_future_wait_for (fut, 30.0) == 0
            && flux_job_event_watch_get (fut, &event) == 0) {
         if (!(o = eventlog_entry_decode (event))) {
             shell_log_errno ("Error decoding eventlog entry");


grondo (approved):
LGTM!

PR #196, opened 2024-08-09
Event watch cancel

### LINKED ISSUE #195: `dws_environment` shell plugin should cancel eventlog watch
The `dws_environment` job shell plugin watches the job eventlog until the `dws_environment` event is found:

https://github.com/flux-framework/flux-coral2/blob/799d36ba5b16752438430973476868c27d86e45f/src/shell/plugins/dws_environment.c#L167-L178

It isn't obvious, but before destroying the future, the plugin should call `flux_job_event_watch_cancel (fut)` to end the streaming RPC, otherwise the watch will stay active in the job-info service, which will continue to send responses as new events appear in the eventlog until the shell exits and a `disconnect` message is sent.

jameshcorbett:
Problem: as described in issue https://github.com/flux-framework/flux-coral2/issues/195, the dws_environment shell
plugin should call flux_job_event_watch_cancel after finding the
dws_environment event, otherwise the watch will stay active in
the job-info service, which will continue to send responses as new
events appear in the eventlog until the shell exits and a
disconnect message is sent.

Call flux_job_event_watch_cancel.

Fixes #195.

### DIFF
--- a/src/shell/plugins/dws_environment.c
+++ b/src/shell/plugins/dws_environment.c
@@ -171,9 +171,12 @@ static int dws_environment_init (flux_plugin_t *p,
     }
     if (read_future (shell, fut) < 0) {
         shell_log_error ("Error reading DW environment from eventlog");
+        flux_job_event_watch_cancel (fut);
         flux_future_destroy (fut);
         return -1;
     }
+    if ((flux_job_event_watch_cancel (fut)) < 0)
+        shell_log_error ("flux_job_event_watch_cancel");
     flux_future_destroy (fut);
     return 0;
 }
--- a/src/shell/plugins/dws_environment.c
+++ b/src/shell/plugins/dws_environment.c
@@ -104,7 +104,13 @@ static int read_future (flux_shell_t *shell, flux_future_t *fut)
             json_decref (o);
             return -1;
         }
-        if (!strcmp (name, "dws_environment")) {
+        if (!strcmp (name, "start")) {
+            //  'start' event with no dws_environment event.
+            shell_log_error ("'start' event found before 'dws_environment'");
+            json_decref (o);
+            return -1;
+        }
+        else if (!strcmp (name, "dws_environment")) {
             if (json_unpack (context,
                              "{s:o, s:o}",
                              "variables",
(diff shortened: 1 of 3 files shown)


src/shell/plugins/dws_environment.c
```diff
@@ -171,11 +172,14 @@ static int dws_environment_init (flux_plugin_t *p,
     }
     if (read_future (shell, fut) < 0) {
         shell_log_error ("Error reading DW environment from eventlog");
+        flux_job_event_watch_cancel (fut);
```
grondo, line 181: I think any error terminates a streaming RPC, so the `flux_job_event_watch_cancel()` is not necessary here.

See [RFC 6](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_6.html#streaming-responses):
> The response stream SHALL consist of zero or more non-error responses, terminated by exactly one error response.

src/shell/plugins/dws_environment.c
```diff
@@ -171,11 +172,14 @@ static int dws_environment_init (flux_plugin_t *p,
...
         flux_future_destroy (fut);
         return -1;
     }
+    if ((rc = flux_job_event_watch_cancel (fut)) < 0)
+        shell_log_error ("flux_job_event_watch_cancel");
     flux_future_destroy (fut);
-    return 0;
+    return rc;
```
grondo, line 188: Though an error here is highly unlikely, I wouldn't have failure of `flux_job_event_watch_cancel()` cause the function to return failure.

grondo (approved):
LGTM! A couple suggestions inline.

src/shell/plugins/dws_environment.c
jameshcorbett: The `read_future` function could return `< 0` for some other reason though, and not because the response stream gave an error response? But I suppose that in any case if the plugin returns `-1` the shell will fail and disconnect...

src/shell/plugins/dws_environment.c
jameshcorbett: Makes sense, fixed and force-pushed.

src/shell/plugins/dws_environment.c
grondo: Ah, good point, up to you then.

jameshcorbett: Since it's not doing any harm I'll leave it. Thanks for the comments! Setting MWP.

PR #202, opened 2024-08-27
Enforce size constraints

jameshcorbett:
Problem: as rabbits enter more general use, there needs to be a way
to enforce crude limits on the size of file systems that users can
request.

Add command-line arguments to coral2_dws for accepting size limits
and logic to the directivebreakdown module for enforcing the limits.

The limits are enforced by adding an exception with a message while the job is held in DEPEND state.

### DIFF
--- a/t/t1002-dws-workflow-obj.t
+++ b/t/t1002-dws-workflow-obj.t
@@ -410,6 +410,73 @@ test_expect_success 'dws service script handles restarts while a job is running'
 	flux job wait-event -vt 25 ${jobid} clean
 '
 
+test_expect_success 'launch service with storage maximum arguments' '
+	flux job cancel $DWS_JOBID &&
+	DWS_JOBID=$(flux submit \
+		--setattr=system.alloc-bypass.R="$R" \
+		-o per-resource.type=node --output=dws4.out --error=dws4.err \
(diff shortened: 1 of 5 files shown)


t/t1002-dws-workflow-obj.t
```diff
@@ -410,6 +410,70 @@ test_expect_success 'dws service script handles restarts while a job is running'
...
+	flux job wait-event -vt 10 ${jobid} exception &&
+	flux job wait-event -vt 15 ${jobid} clean
+'
+
+test_expect_success 'job submission with lustre storage beyond max fails' '
+	jobid=$(flux submit --setattr=system.dw="#DW jobdw capacity=120GiB type=lustre name=project1" \
+		-N1 -n1 hostname) &&
+	flux job wait-event -vt 10 -m description=${CREATE_DEP_NAME} \
+		${jobid} dependency-add &&
+	flux job wait-event -vt 20 ${jobid} exception &&
```
cmoussa1, line 473: Super minor and I'll leave this up to you but it might be nice to check to see that the exception raised on these failed job submissions matches what you expect? i.e `grep` for the exception message raised

cmoussa1 (approved):
LGTM! One tiny nit about a couple of the sharness tests.

t/t1002-dws-workflow-obj.t
jameshcorbett: Good call, done and force-pushed. Will set MWP.

PR #204, opened 2024-08-31
Dws: use static rabbit layout mapping for JGF

### LINKED ISSUE #193: Make `dws2jgf` read from a static file, not k8s
Problem: the `dws2jgf.py` script currently establishes a connection to kubernetes and generates a Fluxion resource graph in JGF format based on the compute-node-to-rabbit mapping it finds there. However, after discussion with admins and other members of the Flux team, it seems that JGF is very difficult to deal with and manipulate (if for instance a queue were to be added) and also can be large enough that it makes a meaningful contribution to the overall size of the ansible repo.

A fix would be to generate the JGF dynamically whenever the management node restarts. However, @trws said that making the resource graph depend on a service (k8s) which may or may not be online is a bad idea. The solution we decided on offline is to keep a rabbit-to-compute-node mapping saved in ansible which `dws2jgf` generates its JGF against. That file could be generated by a script that reads from k8s and periodically updated as needed.

Changes needed:

1. Make a script to go through k8s and write out a rabbit-to-compute-node mapping
2. Change `dws2jgf` to read from the mapping and not k8s

jameshcorbett:
Problem: as described in https://github.com/flux-framework/flux-coral2/issues/193, 
JGF is too unwieldy to be stored in Ansible. On the other hand, Flux's ability to start up and run
jobs cannot be dependent on the responsiveness of kubernetes, so generating JGF from kubernetes 
before starting Flux is not a good option.

The solution I've decided to go with after some discussion is to store some 
static rabbit layout data in ansible, generated by reading from kubernetes. This data in turn is then 
read in to generate JGF. Unlike kubernetes, we can expect the static file to always exist.

Fixes #193

### DIFF
--- a/t/t2000-dws2jgf.t
+++ b/t/t2000-dws2jgf.t
@@ -33,27 +33,32 @@ test_expect_success HAVE_JQ 'smoke test to ensure the storage resources are expe
 	test $(hostname) = compute-01
 '
 
+test_expect_success HAVE_JQ 'flux-rabbitmapping outputs expected mapping' '
+	flux python ${FLUX_SOURCE_DIR}/src/cmd/flux-rabbitmapping.py -i2 > rabbits.json
+	test_cmp ${DATADIR}/rabbits.json rabbits.json
+'
+
 test_expect_success HAVE_JQ 'flux-dws2jgf.py outputs expected JGF for single compute node' '
 	flux R encode -Hcompute-01 | flux python ${CMD} --no-validate --cluster-name=ElCapitan \
-	| jq . > actual-compute-01.jgf &&
+		rabbits.json | jq . > actual-compute-01.jgf &&
 	test_cmp ${DATADIR}/expected-compute-01.jgf actual-compute-01.jgf
 '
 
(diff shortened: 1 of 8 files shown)


t/t2000-dws2jgf.t
```diff
@@ -33,27 +33,32 @@ test_expect_success HAVE_JQ 'smoke test to ensure the storage resources are expe
 	test $(hostname) = compute-01
 '
 
+test_expect_success HAVE_JQ 'flux-rabbitmapping outputs expected mapping' '
```
grondo, line 36: This test has `HAVE_JQ` but I don't see any use of jq in the test itself.

However, also note that flux-core tests and thus `flux-sharness.sh` assume `jq` availability, so at least in flux-core we've dropped use of `HAVE_JQ` and just assume its there (flux-sharness throws an error if not).

(Maybe for future cleanup, no need to address this now)

grondo (approved):
LGTM! Nice improvement. I just noticed one trivial issue in the tests.

PR #206, opened 2024-09-04
testsuite: sharness sync

jameshcorbett:
As @grondo mentioned [here](https://github.com/flux-framework/flux-coral2/pull/204#discussion_r1744383116), we can make `jq` a prereq for the entire testsuite and drop the `HAVE_JQ` prereq from individual test cases.


grondo (approved):
:+1:

PR #208, opened 2024-09-05
testsuite: add a test for big rabbit jobs

jameshcorbett:
Problem: flux-framework/flux-sched/issues/1284 came up in production and was not caught beforehand because the testsuite never exhausts the available rabbit resources.

Add a test that runs three back-to-back 10TiB rabbit jobs to exhaust all the rabbit resources.

### DIFF
--- a/t/t1002-dws-workflow-obj.t
+++ b/t/t1002-dws-workflow-obj.t
@@ -429,8 +429,51 @@ test_expect_success 'dws service script handles restarts while a job is running'
 	flux job wait-event -vt 25 ${jobid} clean
 '
 
+test_expect_success 'back-to-back job submissions with 10TiB file systems works' '
+	jobid1=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=xfs name=project1" \
+		-N1 -n1 hostname) &&
+	jobid2=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=lustre name=project2" \
+		-N1 -n1 hostname) &&
+	jobid3=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=xfs name=project3" \
+		-N1 -n1 hostname) &&
+	flux job wait-event -vt 15 -m description=${CREATE_DEP_NAME} \
+		${jobid1} dependency-add &&
+	flux job wait-event -vt 15 -m description=${CREATE_DEP_NAME} \
+		${jobid2} dependency-add &&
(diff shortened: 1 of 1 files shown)


jameshcorbett:
Dropping WIP because https://github.com/flux-framework/flux-sched/issues/1284 has been resolved.

t/t1002-dws-workflow-obj.t
```diff
@@ -429,8 +429,51 @@ test_expect_success 'dws service script handles restarts while a job is running'
 	flux job wait-event -vt 25 ${jobid} clean
 '
 
+test_expect_success 'back-to-back job submissions with 10TiB file systems works' '
+	jobid1=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=xfs name=project1" \
+		-N1 -n1 hostname) &&
+	jobid2=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=lustre name=project2" \
+		-N1 -n1 hostname) &&
+	jobid3=$(flux submit --setattr=system.dw="#DW jobdw capacity=10TiB type=xfs name=project2" \
```
cmoussa1, line 437: does the name here matching `jobid2`'s name matter here? Maybe it doesn't and you can just ignore this

t/t1002-dws-workflow-obj.t
```diff
@@ -429,8 +429,51 @@ test_expect_success 'dws service script handles restarts while a job is running'
...
+		-N1 -n1 hostname) &&
+	flux job wait-event -vt 15 -m description=${CREATE_DEP_NAME} \
+		${jobid1} dependency-add &&
+	flux job wait-event -vt 15 -m description=${CREATE_DEP_NAME} \
+		${jobid2} dependency-add &&
```
cmoussa1, line 442: Do you also need to check for a `dependency-add` event for `jobid3` here?

cmoussa1 (approved):
LGTM! Just two minor clarifications for the test you added

t/t1002-dws-workflow-obj.t
jameshcorbett: Not strictly necessary but it's a good idea!

jameshcorbett: Done.

jameshcorbett:
Thanks @cmoussa1 ! I addressed your comments and force-pushed, setting MWP.

t/t1002-dws-workflow-obj.t
jameshcorbett: It doesn't matter but it is confusing so I fixed it.

PR #209, opened 2024-09-05
dws2jgf: drop defaults for jgf simplification

jameshcorbett:
Problem: JGF is too large, as described in #193.

As a first step to shrink it, do not output default values, instead skipping them and letting the JGF reader supply them.

Depends on changes introduced in flux-sched by flux-framework/flux-sched/pull/1293. This will remain as a draft until that is merged.


cmoussa1 (approved):
LGTM!!

PR #211, opened 2024-09-16
CI: update from el8 to el9

jameshcorbett:


cmoussa1 (approved):
LGTM! I should probably do the same for flux-accounting

jameshcorbett:
Thanks! Setting MWP

PR #213, opened 2024-09-18
dws: ignore nonexistent rabbit status field

jameshcorbett:
Problem: HPE recently changed the way `Storage` resources have their `status` field populated, so it is no longer guaranteed that the field exists.

Ignore the field if it cannot be found.


cmoussa1 (approved):
LGTM!

PR #215, opened 2024-09-19
dws: only allow owner to create persistent

### LINKED ISSUE #205: restrict persistent rabbit storage to instance owners
@trws said that as a first step for getting rabbits available to users, we should restrict the creation of persistent storage to instance owners, so that admins can control it until we are ready with a better solution.

jameshcorbett:
Problem: until futher work is done to support various usage policies for persistent file systems, it is sometimes desirable to restrict their creation to the instance owner.

Reject jobs with `create_persistent` strings unless the job was submitted by the instance owner.

Add a flag to disable the check.

Fixes #205.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -266,12 +266,21 @@ def move_workflow_desiredstate(workflow_name, desiredstate, k8s_api):
     )
 
 
+def owner_uid(handle):
+    """Get instance owner UID"""
+    try:
+        return int(handle.attr_get("security.owner"))
+    except Exception:
+        return os.getuid()
+
+
 @message_callback_wrapper
-def create_cb(handle, _t, msg, api_instance):
+def create_cb(handle, _t, msg, arg):
     """dws.create RPC callback. Creates a k8s Workflow object for a job.
 
     Triggered when a new job with a jobdw directive is submitted.
     """
+    api_instance, unrestricted_persistent = arg
     dw_directives = msg.payload["dw_directives"]
     jobid = msg.payload["jobid"]
     userid = msg.payload["userid"]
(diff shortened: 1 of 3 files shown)


behlendorf:
Who is meant by the "instance owner" here?  I'm not sure I follow how this is intended to work.

garlick:
It's the userid that is running the flux instance.  For the system instance it's  the `flux` user.
https://flux-framework.readthedocs.io/projects/flux-core/en/latest/guide/glossary.html#term-instance-owner

jameshcorbett:
In theory I could also restrict it to `root` but flux doesn't allow `root` to run jobs so that wouldn't work.

jameshcorbett:
@garlick do you have any tips for how I could submit a job as another user in the testsuite to make sure it's rejected?

grondo:
Note that if you are using test execution then it is a lot less complicated. Just take the `t/scripts/sign-as.py` script from flux-core, and use this function:

```sh
submit_as_alternate_user()
{
        FAKE_USERID=42
        flux run --dry-run "$@" | \
          flux python ${SHARNESS_TEST_SRCDIR}/scripts/sign-as.py $FAKE_USERID \
            >job.signed
        FLUX_HANDLE_USERID=$FAKE_USERID \
          flux job submit --flags=signed job.signed
}
```

See [t2400-job-exec-test.t](https://github.com/flux-framework/flux-core/blob/master/t/t2400-job-exec-test.t) for example of usage. This requires a flux-core compiled with flux-security, so you'll want to use a prereq as in that test as well.

wihobbs (approved):
IMO this implementation tracks with our recent discussions in the Fluxion meetings. Approving.

src/modules/coral2_dws.py
```diff
@@ -284,6 +285,12 @@ def create_cb(handle, _t, msg, api_instance):
         raise TypeError(
             f"Malformed dw_directives, not list or string: {dw_directives!r}"
         )
+    for directive in dw_directives:
+        if not unrestricted_persistent and "create_persistent" in directive:
+            if userid != os.getuid():
```
wihobbs, line 290: If this is running at the system instance level you could call getattr on security.owner. But up to you

jameshcorbett: That's true and might be better, I force-pushed to make it do that instead. Setting MWP.

PR #224, opened 2024-10-01
dws: fix datamovement limit call

jameshcorbett:
Problem: on Tuolumne kubernetes was complaining that it could not interpret `None` as an int in a call to fetch datamovements.

Do not pass `None` as an argument.


cmoussa1 (approved):
LGTM!!

jameshcorbett:
Thanks! Rebased with master, setting MWP.

PR #232, opened 2024-10-28
dws2jgf: remove python 3.7+ feature

jameshcorbett:
Problem: the `capture_output` argument to subprocess.run requires python 3.7+ but LC machines are still on 3.6.

Replace the argument with a 3.6+ compatible equivalent.


cmoussa1 (approved):
LGTM! 👍

PR #233, opened 2024-11-01
dws: use systemconfiguration for rabbit mapping

### LINKED ISSUE #221: Replace some calls to `storage` resources with a single call to `systemconfiguration`
There are some places in this repo where code looks at `storage` resources to figure out which rabbits map to which computes. But some rabbits may be down and therefore would not report which compute nodes they are connected to. This can lead to errors at various points. As @matthew-richerson pointed out, the code could check the `SystemConfiguration` resource instead:

> The SystemConfiguration resource is filled in by someone (or some tool) with how the system is supposed to look. The Storages resources give status on how things currently look. The Storages for the missing Rabbits don't show any computes because the Rabbits aren't there to say which computes they can see

jameshcorbett:
flux-coral2 code in various places reads rabbit info (in particular, the mapping from compute nodes to rabbit nodes and vice versa) from `Storage` resources when it should instead read from the `Systemconfiguration` resource, which is comprehensive and static.

Fixes #221.

### DIFF
--- a/src/cmd/flux-rabbitmapping.py
+++ b/src/cmd/flux-rabbitmapping.py
@@ -7,26 +7,22 @@ import subprocess
 
 import flux
 from flux.hostlist import Hostlist
+from flux_k8s import crd
 
 import kubernetes as k8s
 from kubernetes.client.rest import ApiException
 
 
-def get_storage(config_file):
+def get_k8s_api(config_file):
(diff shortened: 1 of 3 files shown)


cmoussa1 (approved):
LGTM! Nice work

jameshcorbett:
Thanks! Setting MWP.

PR #235, opened 2024-11-05
Rabbit preset dw strings

### LINKED ISSUE #200: Add preset `run/batch/submit/alloc` flags for simplifying rabbit submission
The normal way to ask for rabbit storage for a job is to add the `--setattr=dw="#DW jobdw type=xfs capacity=100GiB name=myname"` option to `flux submit/run/batch/alloc` but I was discussing today with @behlendorf whether there could be some simpler preset flags with configurable meanings. What we discussed was something like have flags like `--small-xfs` or `--large-lustre` that simply add a `#DW` string to the user's job, where that string could be configured at the system level, maybe in a config file somewhere. This would just be to simplify things for most users, advanced users could still set the `dw` attribute directly.

Thoughts @garlick @grondo ?

jameshcorbett:
Problem: some users may not want to write custom #DW strings for
their jobs. It can take a while to learn the syntax and figure out
what options to use.

Add configurable preset #DW strings that can be defined in a config
file. Administrators can define some presets they think would work
for a large number of users of the system.

Fixes #200 .

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -282,16 +282,23 @@ def create_cb(handle, _t, msg, arg):
     dw_directives = msg.payload["dw_directives"]
     jobid = msg.payload["jobid"]
     userid = msg.payload["userid"]
+    presets = handle.conf_get("rabbit.presets", {})
     if isinstance(dw_directives, str):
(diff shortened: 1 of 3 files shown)


cmoussa1 (approved):
LGTM!

Issue #197, opened 2024-08-10
DWS: handle broken kubernetes connection


jameshcorbett:
An error occurred on elcap:

```
Traceback (most recent call last):
  File "/usr/bin/coral2_dws.py", line 937, in <module>
    main()
  File "/usr/bin/coral2_dws.py", line 928, in main
    handle.reactor_run()
  File "/usr/lib64/flux/python3.6/flux/core/handle.py", line 322, in reactor_run
    Flux.raise_if_exception()
  File "/usr/lib64/flux/python3.6/flux/core/handle.py", line 133, in raise_if_exception
    raise cls.set_exception(None) from None
  File "/usr/lib64/flux/python3.6/flux/core/watchers.py", line 68, in timeout_handler_wrapper
    watcher.callback(watcher.flux_handle, watcher, revents, watcher.args)
  File "/usr/lib64/flux/python3.6/flux_k8s/watch.py", line 63, in watch_cb
    watchers.watch()
  File "/usr/lib64/flux/python3.6/flux_k8s/watch.py", line 99, in watch
    watch.watch()
  File "/usr/lib64/flux/python3.6/flux_k8s/watch.py", line 47, in watch
    for event in stream:
  File "/usr/lib/python3.6/site-packages/kubernetes/watch/watch.py", line 144, in stream
    for line in iter_resp_lines(resp):
  File "/usr/lib/python3.6/site-packages/kubernetes/watch/watch.py", line 46, in iter_resp_lines
    for seg in resp.read_chunked(decode_content=False):
  File "/usr/lib/python3.6/site-packages/urllib3/response.py", line 694, in read_chunked
    self._original_response.close()
  File "/usr/lib64/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/lib/python3.6/site-packages/urllib3/response.py", line 378, in _error_catcher
    raise ProtocolError('Connection broken: %r' % e, e)
  urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(2062 bytes read, 608 more expected
```

Relates to #145 and #159.

Because of this error a bunch of workflows were stranded in PreRun because the coral2-dws service was down when the `dws.post_run` RPCs were sent. And perhaps relatedly, it seems that a lot of `dws.prolog_remove` RPCs were sent (perhaps because the workflows had some updates occur while `PreRun` was `ready: True`) because @grondo noted that the Flux logs were filling up with messages like 

```
[...]

Issue #159, opened 2024-05-10
Configure coral2-dws's k8s connections to tolerate more lost connections


jameshcorbett:
Problem: the coral2-dws service on elcap sometimes loses connection to the k8s server, logging

```
WARNING - Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))': /apis/dataworkflowservices.github.io/v1alpha2/namespaces/default/workflows?resourceVersion=0&timeoutSeconds=1&watch=True
```

Sometimes this can cause workflows to become stuck.

Somehow the service should become more resilient, and keep retrying with a backoff.

roehrich-hpe:
We think we have a fix for this in https://github.com/flux-framework/flux-coral2/issues/165

PR #237, opened 2024-11-08
dws: retry k8s failures while cleaning up workflows

jameshcorbett:
Related to #159 and #197 . Add async routines to try to teardown and delete k8s workflows in an infinite loop until success.


cmoussa1 (approved):
LGTM!!

PR #241, opened 2024-12-04
dws: fix NameError in cleanup.py

jameshcorbett:
Problem: cleanup.py refers to the `ConfigException` type but doesn't import it from the proper kubernetes module.

Import it, and fix some other problems pylint identified.


cmoussa1 (approved):
LGTM!

jameshcorbett:
Thanks!

PR #245, opened 2024-12-16
Add check for no DWS Storages

bdevcich:
Problem: When there are no `Storages` available in k8s, no list of rabbits is available. This essentially results in a KeyError since the rabbit dicts are empty:

```
2024-12-16 10:06:18,840 - ERROR - Error in responding to dws.setup RPC for 14489491083362304: 'rabbit-compute-2'
```

Add a check to make sure there are `Storages` available, if not, then exit since that means there are no rabbits available, and no workflows will be able to run.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -172,7 +172,7 @@ def get_datamovements(k8s_api, workflow_name, count):
     try:
         api_response = k8s_api.list_cluster_custom_object(
             group="nnf.cray.hpe.com",
-            version="v1alpha2",
+            version="v1alpha4",
             plural="nnfdatamovements",
             label_selector=(
                 f"dataworkflowservices.github.io/workflow.name={workflow_name},"


jameshcorbett:
@bdevcich how come you closed this?

bdevcich:
> @bdevcich how come you closed this?

My fork of the repo was very old when I made this fix and I didn't realize that until after I already made the PR. Once I looked at the latest code, it seems like this would no longer be an issue - unless I am mistaken.

jameshcorbett:
Ah I can't see what your changes were so I just wanted to be sure. I think it shouldn't be an issue now that the `systemconfiguration` is used instead of `storages` to populate the maps.

bdevcich:
> I think it shouldn't be an issue now that the `systemconfiguration` is used instead of `storages` to populate the maps.

This is exactly what I thought when I pulled in the latest code, but wasn't able to confirm it on our systems.

Issue #158, opened 2024-05-10
Rabbit resources jobspec update


jameshcorbett:
The dws-jobtap plugin grabs a reference to the user's jobspec and modifies it [in an RPC callback](https://github.com/flux-framework/flux-coral2/blob/fa383fff9d3c333a9a3a7e5fc40dcde150a0a78b/src/job-manager/plugins/dws-jobtap.c#L585).

The flux-core `jobtap` interface exposes a function `flux_jobtap_jobspec_update_pack (flux_plugin_t *p, const char *fmt, ...)`. The dws-jobtap plugin should use that function (since it is part of the official interface) rather than relying on manipulation of the copy of jobspec held by the job manager. However, the function does not accept a `jobid` argument, which it would need to do in order to be used properly by `dws-jobtap`, since `dws-jobtap` does not modify the jobspec from within a jobtap callback.

PR #250, opened 2025-01-14
dws: rabbit resources jobspec update

jameshcorbett:
Problem: The dws-jobtap plugin does not use the jobspec update
jobtap interface to perform its update of a job's resources. Instead
it simply grabs the jobspec directly and manipulates it. However, it
would be good for users to have a record of how their jobspec was
changed, and doing so may prevent problems in the future.

Make dws-jobtap use the jobspec update interface.

Fixes https://github.com/flux-framework/flux-coral2/issues/158


grondo (approved):
LGTM!

PR #251, opened 2025-01-15
test: container image updates

jameshcorbett:
flux-sched no longer supports the `focal` image, so replace it with `noble`.


grondo (approved):
LGTM!

PR #252, opened 2025-01-16
dws: add negative requires hostlist for down rabbits

jameshcorbett:
Problem: Fluxion continues to have problems with rabbits on
production systems. However, users need a way for Flux to
automatically schedule around down rabbits.

As a workaround, maintain a hostlist of all nodes that are attached
to down rabbits. Add this hostlist to jobs' constraints, as a negative hostlist constraint (like
`--requires=-hosts:foobar[1-20]`)

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -556,6 +556,61 @@ static int exception_cb (flux_plugin_t *p,
     return 0;
 }
 
+/*
+ * Generate a new jobspec constraints object for a job so that it can avoid
+ * attempting to run on nodes attached to down rabbits.
+ */
+static json_t *generate_constraints(flux_t *h, flux_plugin_t *p, flux_jobid_t jobid, const char *exclude_str){
+    flux_plugin_arg_t *args = flux_jobtap_job_lookup (p, jobid);
+    json_t *constraints = NULL;
+    json_t *not;
+    if (!args
+        || flux_plugin_arg_unpack (args,
+                                   FLUX_PLUGIN_ARG_IN,
+                                   "{s:{s:{s:{s?o}}}}",
+                                   "jobspec",
+                                   "attributes",
+                                   "system",
+                                   "constraints",
+                                   &constraints)
+        < 0) {
+        flux_log_error (h, "Failed to unpack args");
+        flux_plugin_arg_destroy (args);
+        return NULL;
+    }
+    if (!constraints) {
+        if (!(constraints = json_pack ("{s:[{s:[s]}]}", "not", "hostlist", exclude_str))){
+            flux_log_error (h, "Failed to create new constraints object");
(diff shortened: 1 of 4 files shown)


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -556,44 +556,111 @@ static int exception_cb (flux_plugin_t *p,
     return 0;
 }
 
+/*
+ * Generate a new jobspec constraints object for a job so that it can avoid
+ * attempting to run on nodes attached to down rabbits.
+ */
+static json_t *generate_constraints(flux_t *h, flux_plugin_t *p, flux_jobid_t jobid, const char *exclude_str){
+    flux_plugin_arg_t *args = flux_jobtap_job_lookup (p, jobid);
```
cmoussa1, line 564: Just curious: do you need to call `flux_plugin_arg_destroy ()`  once you are done with `args`? I needed to do this for my C++ jobtap plugin that does something similar in order to get my tests to pass under valgrind. Maybe this is not needed in a C jobtap plugin though, and if so you should just ignore me

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -556,44 +556,111 @@ static int exception_cb (flux_plugin_t *p,
...
+    json_t *not;
+    if (!args
+        || flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:{s:{s:{s?o}}}}",
+                                "jobspec",
+                                "attributes",
+                                "system",
+                                "constraints",
+                                &constraints)
```
cmoussa1, line 575: perhaps indent these params to line up with `args`?

cmoussa1 (approved):
LGTM! 👍 One minor clarification question

src/job-manager/plugins/dws-jobtap.c
grondo: Good catch @cmoussa1! Here's the comment in `jobtap.h` for this function:
```c
/*  Return a flux_plugin_arg_t object for a job.
 *
 *  The result can then be unpacked with flux_plugin_arg_unpack(3) to get
 *   active job information such as userid, state, etc.
 *
 *  If `id` is not an active job or is not found in the job manager's
 *   inactive job cache then NULL is returned with ENOENT set.
 *
 *  Caller must free with flux_plugin_arg_destroy(3).
 */
flux_plugin_arg_t * flux_jobtap_job_lookup (flux_plugin_t *p,
                                            flux_jobid_t id);

```

jameshcorbett: Thanks both of you! Fixed. @cmoussa1 want to give another pass or should I set MWP?

cmoussa1: LGTM! Feel free to set MWP

PR #254, opened 2025-01-25
pals: make eventlog timeout configurable

jameshcorbett:
Problem: sometimes 10 seconds is not enough of a timeout for reading from the eventlog if lustre is hanging.

Make the timeout configurable.

### DIFF
--- a/src/shell/plugins/cray_pals.c
+++ b/src/shell/plugins/cray_pals.c
@@ -490,7 +490,8 @@ error:
 static int read_future (flux_future_t *fut,
                         char *buf,
                         size_t bufsize,
-                        json_int_t *random)
+                        json_int_t *random,
+                        double timeout)
 {
     json_t *o = NULL;
     json_t *context = NULL;
@@ -501,7 +502,7 @@ static int read_future (flux_future_t *fut,
     json_int_t portnum;
     int bytes_written;
 
-    while (flux_future_wait_for (fut, 10.0) == 0
+    while (flux_future_wait_for (fut, timeout) == 0
            && flux_job_event_watch_get (fut, &event) == 0) {
         if (!(o = eventlog_entry_decode (event))) {
             shell_log_errno ("Error decoding eventlog entry");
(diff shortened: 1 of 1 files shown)


grondo (approved):
This looks reasonable to me!

Stepping back a bit though, I'm still not sure exactly what went wrong in the failures we saw recently. If I remember correctly how this is supposed to work, the pals shell plugin is reading the job eventlog in order to fetch the `cray_port_distribution`  event (and only that?), which is emitted from a  jobtap plugin prolog.

Since the job shells are not started until after all `prolog-finish` events, the `cray_port_distribution` event should be guaranteed to be in the eventlog before the first shell is started (modulo the batch commit interval in the job-manager), so I'm still confused on why job shells were stuck waiting for that event (and looking back at the job in question, event the job `start` event appears in the eventlog)

Do you recall if we ever narrowed down why the eventlog read could time out in this situation?

I guess as another comment, it would be nice to know if the eventlog watch returned _any_ events, or if it was stuck after reading a particular event. To that end, maybe a second commit here could amend the error message in this function when it times out to include the last event (if any) that was seen in the eventlog. I guess that may help us determine if the call received any response at all (my guess is not).

jameshcorbett:
> Do you recall if we ever narrowed down why the eventlog read could time out in this situation?

No, I don't think so. 

> I guess as another comment, it would be nice to know if the eventlog watch returned any events, or if it was stuck after reading a particular event. To that end, maybe a second commit here could amend the error message in this function when it times out to include the last event (if any) that was seen in the eventlog. I guess that may help us determine if the call received any response at all (my guess is not).

Ok I'll work on that!

> Stepping back a bit though, I'm still not sure exactly what went wrong in the failures we saw recently. If I remember correctly how this is supposed to work, the pals shell plugin is reading the job eventlog in order to fetch the cray_port_distribution event (and only that?), which is emitted from a jobtap plugin prolog.

Yeah, that's right.

jameshcorbett:
> > I guess as another comment, it would be nice to know if the eventlog watch returned any events, or if it was stuck after reading a particular event. To that end, maybe a second commit here could amend the error message in this function when it times out to include the last event (if any) that was seen in the eventlog. I guess that may help us determine if the call received any response at all (my guess is not).
> 
> Ok I'll work on that!

Done, can you review again @grondo ?

grondo (approved):
LGTM!

PR #259, opened 2025-01-30
dws: add 'mapping' to accepted config keys

jameshcorbett:
Problem: 'rabbit.mapping' is rejected by coral2_dws as a config key, but it is valid.

Make coral2_dws accept it.


cmoussa1 (approved):
LGTM!

PR #260, opened 2025-02-04
dws: make config error a warning

jameshcorbett:
Problem: if unexpected keys are found in the `rabbit` config table, coral2_dws exits with an error. This is probably overkill.

Instead, issue a warning.


cmoussa1 (approved):
LGTM!

PR #263, opened 2025-02-06
dws: cancel data movement, move workflow to Teardown

### LINKED ISSUE #255: Add a way to force-cancel rabbit data movement
After a rabbit job finishes, an epilog is placed on them while data movement completes, because the job should not be considered finished until data has been moved. However, that epilog also prevents the jobs from being canceled and moving to INACTIVE. This is intentional because I thought users might not want to destroy their data just because they canceled their job.

@bdevcich has found that sometimes workflows get stuck in the `DataOut` stage because DCP hangs. When that happens, it requires manual intervention from an admin with kubernetes write access, to move the workflow manually to the `Teardown` state. That is bad.

I see two options here, to ensure manual intervention is never needed:

1. If a `cancel` event is received while the job is in the data movement epilog, somehow interpret that as meaning "forget data movement, just get rid of the job" and move the workflow to `Teardown`.
2. Add a separate utility, like `flux rabbit-cancel`, that moves a workflow to `Teardown`.

@grondo do you have any thoughts?

jameshcorbett:
Problem: coral2_dws does not have a way to make a workflow jump to
Teardown, which may sometimes be necessary.

Add an RPC that moves workflows to Teardown.

Fixes #255.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -477,7 +477,12 @@ static int cleanup_cb (flux_plugin_t *p,
                                      "flux::dws_run_started")) {
             dws_run_started = 1;
         }
-        if (flux_jobtap_epilog_start (p, DWS_EPILOG_NAME) < 0) {
+        if (flux_jobtap_job_aux_set (p,
+                                     id,
+                                     "dws_epilog_active",
+                                     (void*) 1,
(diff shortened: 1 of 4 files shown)


cmoussa1 (approved):
LGTM!! 👍

bdevcich:
Thanks @jameshcorbett!

PR #264, opened 2025-02-06
`getrabbit`: list all rabbits

jameshcorbett:
Problem: there is no way to get a list of all rabbit nodes from
the `flux getrabbit` command.

Make it so that passing no arguments to `getrabbit` makes it print
a hostlist of all rabbits.


cmoussa1 (approved):
LGTM!! Looks really useful :-)

PR #265, opened 2025-02-07
dws: change exclusion constraint to a property 

jameshcorbett:
Problem: Fluxion continues to have problems with rabbits on
production systems. However, users need a way for Flux to
automatically schedule around down rabbits.

Previously (see #252), a hostlist constraint was used. However, that doesn't
work very well when queue times are long, because rabbit status can
change while the job sits in the queue.

Instead, use a property constraint.


cmoussa1 (approved):
LGTM! Nice improvements to the sharness tests too 👍

PR #266, opened 2025-02-07
dws: add option to tag compute nodes instead of draining

jameshcorbett:
Problem: in theory, if a rabbit-to-compute-node PCIe link goes down,
coral2_dws will drain the node. However, admins have generally
disabled that feature for being disruptive to non-rabbit users.
PR https://github.com/flux-framework/flux-coral2/pull/265 introduced a non-intrusive way to avoid scheduling rabbit
jobs on specific nodes, by tagging nodes with a property. This
could be used to cover the PCIe link failure case as well.

If draining is disabled, unless the `rabbit.soft_drain` config
option is set to `false`, mark nodes with a failed PCIe link
with the `badrabbit` property.


cmoussa1 (approved):
LGTM!

PR #269, opened 2025-02-12
dws: remove-property fix

jameshcorbett:
Fixes two separate but related issues, described in the first two commit messages.


cmoussa1 (approved):
LGTM!

PR #272, opened 2025-02-13
rabbitmapping: fetch by jobid

### LINKED ISSUE #268: Make `flux getrabbit` accept jobids
Problem: some users want to know what rabbits their jobs are attached to. Currently this is possible with ``flux getrabbit -c $(flux jobs -no "{nodelist}" $JOBID)`` but it should be simple enough to add a `--jobid` flag.

jameshcorbett:
Problem: `flux getrabbit` could save users a bit of work by accepting
job IDs.

Make it take job IDs.

Closes #268 .

### DIFF
--- a/src/cmd/flux-getrabbit.py
+++ b/src/cmd/flux-getrabbit.py
@@ -11,13 +11,8 @@ from flux.job import JobID
 from flux.hostlist import Hostlist
 
 
-def main():
-    """Create a JSON file mapping compute nodes <-> rabbits.
-
-    Fetch the SystemConfiguration from kubernetes and use that for the mapping.
-    Also fetch Storage resources from kubernetes to populate the JSON file with
-    capacity data.
-    """
+def read_args():
+    """Read in command-line args."""
     parser = argparse.ArgumentParser(
         formatter_class=flux.util.help_formatter(),
         description=("Map compute nodes to rabbits and vice versa."),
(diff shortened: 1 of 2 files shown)


src/cmd/flux-getrabbit.py
```diff
@@ -60,11 +66,22 @@ def main():
...
         print(hlist.uniq().encode())
         return
+    if args.jobids:
+        for jobid in args.jobids:
+            try:
+                JobID(jobid)
+            except Exception as exc:
+                sys.exit(f"Could not interpret {jobid} as a flux Jobid: {exc}")
+            try:
+                nodelist = flux.job.job_list_id(handle, jobid, ["nodelist"]).nodelist
```
grondo: I would assume you need to use `.get_jobinfo()` on the result of `job_list_id()` (a `JobListIdRPC`) to then get the `nodelist` attribute (but I may be missing something obvious).

Also, not critical, but you can get the nodelist for all jobids using one RPC via the `JobList` class. For example, something like:
```python
  jobs = JobList(handle, ids=jobids, attrs=["nodelist"]).jobs()
  for job in jobs:
    args.computes.append(job.nodelist)
```

src/cmd/flux-getrabbit.py
```diff
@@ -28,23 +23,63 @@ def main():
...
+    if not args.computes:
+        args.computes = []
+    if args.jobids:
+        for jobid in args.jobids:
+            try:
+                nodelist = (
+                    flux.job.job_list_id(handle, JobID(jobid), ["nodelist"])
+                    .get_jobinfo()
+                    .nodelist
+                )
```
grondo, line 61: Oh, you fixed my previous comment here :-)

grondo (approved):
LGTM, just one take-it-or-leave-it comment inline.

It might be good to fix the usage of `job_list_id` in the first commit, but that's at your discretion.

src/cmd/flux-getrabbit.py
jameshcorbett: No worries, it's my fault for force-pushing so much. And I probably should have kept the history better anyway. I'm going to set MWP though. The single RPC makes sense, but I imagine 99% of users will just pass one job ID, and it seemed like `JobList()` when I tried it just now failed silently on a bad job ID, whereas `job_list_id` raised an exception, and that may make for a better user experience.

grondo: Eh, I was wrong anyway, `JobList()` uses `job_list_id` under the covers when multiple, specific jobids are passed anyway.  (it just sends the RPCs all at once and returns a single future which is fulfilled when all RPCs are complete).

And agreed that the error handling is awkward (putting it politely) (errors are held in `self.errors`, which you have to check if less jobs were returned than expected)

Anyhow, sounds like you made the right choice here.

PR #277, opened 2025-02-13
`getrabbit`: check system config

jameshcorbett:


cmoussa1 (approved):
LGTM! Is there an easy way to replicate the scenario where `path` is `None`? If so, it might be nice to include in the testsuite.

jameshcorbett:
> LGTM! Is there an easy way to replicate the scenario where `path` is `None`? If so, it might be nice to include in the testsuite.

Good idea, thanks! I force-pushed to add a test for that.

Setting MWP.

PR #279, opened 2025-02-18
`getrabbit`: jobid memo lookups

### LINKED ISSUE #275: `getrabbit`: fetch rabbits for a job from memo, not from static file
Problem: #272 added a way to look up the rabbits for a job. It works correctly in all cases, however in the future for lustre file systems, the rabbits associated with a job may no longer be the rabbits local to the compute nodes in the job. 

There is already [a memo](https://github.com/flux-framework/flux-coral2/blob/8abdb50447589c20985777da9347ffb2db0891ba/src/modules/coral2_dws.py#L319) associated with every rabbit job, specifying what rabbits are used. That should be used instead.

jameshcorbett:
Problem: https://github.com/flux-framework/flux-coral2/pull/272 added a way to look up the rabbits for a job. It works correctly in all cases, however in the future for lustre file systems, the rabbits associated with a job may no longer be the rabbits local to the compute nodes in the job.

There is already [a memo](https://github.com/flux-framework/flux-coral2/blob/8abdb50447589c20985777da9347ffb2db0891ba/src/modules/coral2_dws.py#L319) associated with every rabbit job, specifying what rabbits are used. That should be used instead.

Fixes #275 .

### DIFF
--- a/src/cmd/flux-getrabbit.py
+++ b/src/cmd/flux-getrabbit.py
@@ -55,15 +55,19 @@ def map_to_rabbits(args, handle, mapping):
         for jobid in args.jobids:
             try:
                 nodelist = (
-                    flux.job.job_list_id(handle, JobID(jobid), ["nodelist"])
+                    flux.job.job_list_id(handle, JobID(jobid), ["annotations"])
                     .get_jobinfo()
-                    .nodelist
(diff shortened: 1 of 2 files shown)


cmoussa1 (approved):
LGTM!

Issue #267, opened 2025-02-10
Update cray-pals version


jameshcorbett:
Problem: the cray-pals plugin uses an old version of the apinfo file. It should be updated to something more recent.

garlick:
Yeah, we're on version 1 and slurm's on version 5:

https://github.com/SchedMD/slurm/blob/master/src/plugins/mpi/cray_shasta/apinfo.h#L47-L50

garlick:
Suggestion: to get us started and let us reuse as much code as possible, could HPE submit a pull request to just shovel the whole slurm cray_shasta plugin (which they hold (C) on)  into this repo under `src/common/libpals` and change the license to LGPL?

jameshcorbett:
I emailed someone at HPE about it, they said they could copy the code over for us, but they'd have to go through some release process to do so. They offered to explain the changes between version 1 and 5 instead and I agreed. The following is what they sent me:

Here are the changes made between version 1 and 5.
 
The header structure now looks like this:
```C
typedef struct {
    int version;
    size_t total_size;
    size_t comm_profile_size;
    size_t comm_profile_offset;
    int ncomm_profiles;
    size_t cmd_size;
    size_t cmd_offset;
    int ncmds;
    size_t pe_size;
    size_t pe_offset;
    int npes;
    size_t node_size;
    size_t node_offset;
    int nnodes;
    size_t nic_size;
    size_t nic_offset;
    int nnics;
    size_t status_offset;
    size_t dist_size;
    size_t dist_offset;
} pals_header_t;
```
The status_offset, dist_size, and dist_offset fields are new.
 
The status_offset field gives a way for the launcher to tell if all ranks called pals_init/pals_fini.
To support this feature, write an array of ints all initialized to 0 at the end of the apinfo file, one per local rank.
Then set status_offset to the offset of the start of the array, and increase total_size accordingly.
Each rank will write 1 to its offset in the array when it calls pals_init, and write 2 to its offset when it calls pals_fini.
To disable this feature (if you don't need this information), set status_offset to 0.
 
The dist_size and dist_offset fields are for providing the distances from each rank to each NIC on the node, to give PMI a hint for which NIC to use.
This is optional – set dist_size and dist_offset to 0 to disable.
```C
typedef struct {
    uint8_t num_nic_distances;        /**< Number of CPU->NIC distances */
    uint8_t accelerator_distances;    /**< Accel distances too? (bool) */
[...]

PR #282, opened 2025-02-20
pals: update pals apinfo version to 5

jameshcorbett:
Problem: the cray-pals plugin uses an old version of the apinfo file.

Update it to version 5.

Preliminary work on #267.

### DIFF
--- a/src/shell/plugins/cray_pals.c
+++ b/src/shell/plugins/cray_pals.c
@@ -73,41 +73,39 @@
  */
 
 /* Application file format version */
-#define PALS_APINFO_VERSION 1
+#define PALS_APINFO_VERSION 5
 
-/* File header structure */
 typedef struct {
-    int version;               // Set to PALS_APINFO_VERSION
-    size_t total_size;         // Size of the whole file in bytes
-    size_t comm_profile_size;  // sizeof(pals_comm_profile_t)
-    // offset from beginning of file to the first comm_profile_t
-    size_t comm_profile_offset;
-    // number of comm_profile_t (not used yet, set to 0)
-    int ncomm_profiles;
-    size_t cmd_size;  // sizeof(pals_cmd_t)
-    // offset from beginning of file to the first pals_cmd_t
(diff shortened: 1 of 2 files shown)


garlick:
Nice!

Could we perhaps improve this a bit by creating a convenience library that has its own libtap tests, so that the PALS interface can be tested independently of the shell plugin?

Also, maybe we could extract the known good v1 code, and then add v5 separately and provide a way to select one or the other so there are options if new development introduces a problem?

I'd be willing to submit a PR to do that refactoring.  I had already started one a few weeks ago actually.

garlick:
Apologies, I seem to have a few other high priority things piled up at this point that will consume at least the beginning of the week.   Let me just describe briefly what I had in mind and see how it sounds to you @jameshcorbett:

Step 1: refactor the apinfo version 1 code

Create a standalone class in `src/common/libapinfo`.
- copy the HPE provided struct definitions to `apinfo1.h`.
- a `create` method allocates the apinfo object and populates it with values sensible for a singleton
- a `put` method writes it to a file.  Maybe use a passed in `FILE *` so caller handles open/close
- a `get` method reads it back in again (skip if this is not useful)
- setters allow other aspects of the object to be set
- getters as needed for use in the shell plugin and tests

By standalone, I mean keep the details of the shell plugin interface out.

Develop a libtap test for v1.

Step 2: implement version 5 without stepping on version 1

- Create a new `apinfo5.h`.
- Modify the `apinfo` class to allow create of either version
- probably add more accessors

Develop a libtap test for v5.

So again I think the main benefits of going this way are
- improved opportunities for unit testing
- provide a way to fall back to v1 via shell plugin args (TBD)
- isolation of HPE specific stuff from very Flux specific stuff to encourage engagement

jameshcorbett:
Sounds reasonable to me, but I'll have to circle back again when I have time.

jameshcorbett:
Closing this because it would need a complete refactoring to work with the new apinfo layout.

PR #284, opened 2025-02-21
dws: add jobid to jobtap error messages

jameshcorbett:
Problem: often, the dws-jobtap plugin does not report what jobid it hits an error on. This can make debugging hard.

Add job ids to the error messages.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -35,6 +35,21 @@ struct create_arg_t {
     flux_jobid_t id;
 };
 
+/*  Convenience function to convert a flux_jobid_t to F58 encoding
+ *  If the encode fails (unlikely), then the decimal encoding is returned.
+ */
+static inline const char *idf58 (flux_jobid_t id)
+{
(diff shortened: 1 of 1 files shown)


grondo (approved):
Looks reasonable. If you want f58 encoded jobids, which are simpler to grep for and would be consistent with jobids logged in places in Flux, you could steal `src/common/libjob/idf58.h` from flux-core, which basically uses this:
```c
/*  Convenience function to convert a flux_jobid_t to F58 encoding
 *  If the encode fails (unlikely), then the decimal encoding is returned.
 */
static inline const char *idf58 (flux_jobid_t id)
{
    static __thread char buf[21];
    if (flux_job_id_encode (id, "f58", buf, sizeof (buf)) < 0) {
        /* 64bit integer is guaranteed to fit in 21 bytes
         * floor(log(2^64-1)/log(1)) + 1 = 20
         */
        (void) sprintf (buf, "%ju", (uintmax_t) id);
    }
    return buf;
}
```

jameshcorbett:
Good idea, I'll change the code everywhere to use that.

PR #286, opened 2025-02-24
dws-jobtap: avoid setting aux too early

### LINKED ISSUE #285: `dws-jobtap.c`: prolog can't be removed after flux restart
On elcap, there were a number of jobs stuck with eventlogs like the following:

```
[root@elcap1:corbett8]# flux job eventlog fdPP6KpTScF
1740076466.590968 submit userid=35619 urgency=4 flags=0 version=1
1740076466.678249 jobspec-update attributes.system.bank="guests" attributes.system.project="*"
1740076466.678308 validate
1740076466.740571 dependency-add description="dws-create"
1740076466.752246 memo rabbit_workflow="fluxjob-270261880771618816"
1740076469.806363 jobspec-update resources=[{"type":"node","count":16,"exclusive":true,"with":[{"type":"slot","count":1,"with":[{"type":"core","count":1}],"label":"task"}]}]
1740076469.806509 dependency-remove description="dws-create"
1740076469.806543 depend
1740076469.806627 priority priority=3937
1740078090.456939 priority priority=3622
1740083488.730144 priority priority=3148
1740085290.636922 priority priority=3306
1740087085.199223 priority priority=3148
1740118109.455611 flux-restart
1740118112.231086 jobspec-update attributes.system.bank="guests"
1740118112.231135 priority priority=3148
1740118471.500172 flux-restart
1740118474.054093 jobspec-update attributes.system.bank="guests"
1740118474.054197 priority priority=3148
1740136770.020332 alloc annotations={"user":{"rabbit_workflow":"fluxjob-270261880771618816"}}
1740136770.021013 prolog-start description="job-manager.prolog"
1740136770.021050 prolog-start description="cray-pals-port-distributor"
1740136770.021056 prolog-start description="dws-setup"
1740136770.023734 cray_port_distribution ports=[11925,11924] random_integer=-3707311454147272688
1740136770.023775 prolog-finish description="cray-pals-port-distributor" status=0
1740137999.790855 prolog-finish description="job-manager.prolog" status=0
1740171685.212276 exception type="cancel" severity=0 note="" userid=35619
1740171700.052267 exception type="cancel" severity=0 note="" userid=35619
```

I think what was happening was that the `dws_prolog_active` jobtap job aux was set in the `depend_cb` callback which is called in the `DEPEND` state. Then flux was restarted, and later the aux was looked for in the `RUN` callback. However, the aux was removed by the restart, and so it wasn't found. This is consistent with the following messages in `flux dmesg`:

```
[ +33.512372] job-manager.err[0]: jobtap: job.state.run: callback returned error
[ +33.513515] job-manager.err[0]: Failed to start jobtap prolog for dws: No such file or directory
```

An improvement would be to set the aux only when it's actually needed, in the `RUN` callback. However, that would still leave the plugin vulnerable to flux restarts, it would just narrow the window.

I wonder if there is a better way to handle it though. I'll need to refresh my understanding of what the aux is doing.

jameshcorbett:
Problem: as described in #285, the `dws-jobtap` plugin creates a boolean jobtap aux to indicate whether or not the prolog it sets on jobs is still active, but it creates the aux in the DEPEND state despite the fact that it will not start the prolog until the RUN state. If Flux is restarted with jobs in the queue, the aux will be lost and a lookup for the aux will cause the RUN callback to fail.

Only set the `prolog_active` aux at the time that the prolog is actually started, and not before.

Fixes #285.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -160,7 +160,6 @@ static int depend_cb (flux_plugin_t *p, const char *topic, flux_plugin_arg_t *ar
     json_t *resources;
     json_t *jobspec;
     int userid;
-    int *prolog_active = NULL;
 
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

PR #288, opened 2025-02-25
dws: stop casting k8s resourceVersion to int

jameshcorbett:
Problem: kubernetes documentation states that objects' resourceVersion fields should NOT be cast to an integer and compared. However, coral2_dws does exactly that.

Remove the cast and comparison, since it is not needed.


cmoussa1 (approved):
Nice!

PR #298, opened 2025-02-27
dws: drop unnecessary log messages

jameshcorbett:
`coral2_dws` logs some very verbose messages which clutter the logs, and are unhelpful for debugging. It also includes kubernetes log messages when it attempts to set its own log level to `LOG_DEBUG`.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -53,7 +53,7 @@ def log_rpc_response(rpc):
     except Exception as exc:
         LOGGER.warning("RPC error %s", repr(exc))
     else:
-        if msg is not None:
+        if msg:
             LOGGER.debug("RPC response was %s", msg)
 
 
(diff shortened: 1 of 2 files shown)


wihobbs (approved):
Looks good. Is the log format consistent with our other things in /var/log?

jameshcorbett:
> Looks good. Is the log format consistent with our other things in /var/log?

not that I know of, do you know where I could find examples of that format?

wihobbs:
This is what I was thinking of: 
```
[root@mutt5:hobbs17]# logger -s -t epilog "script: example message"
<13>Feb 27 15:06:13 epilog: script: example message
[root@mutt5:hobbs17]# cat /var/log/epilog.log
2025-02-27T15:06:13.859950-08:00 mutt5 epilog[4080912]: script: example message
```

I think the number after epilog is a pid.

jameshcorbett:
It's similar, the logs look like:

```
Feb 26 21:30:25 elcap1 flux-coral2-dws[1983722]: 2025-02-26 21:30:25,816 - INFO - Found nnfdatamovement crd for workflow 'fluxjob-279214303948240896': {"apiVersion": "nnf.cray.hpe.com/v>
Feb 26 21:30:25 elcap1 flux-coral2-dws[1983722]: 2025-02-26 21:30:25,816 - INFO - Found nnfdatamovement crd for workflow 'fluxjob-279214303948240896': {"apiVersion": "nnf.cray.hpe.com/v>
```

which makes me think that I should drop the `%(asctime)s` field because `journalctl` provides it by default already.

jameshcorbett:
Ok force-pushed to make that change, will set MWP.

wihobbs:
Thanks @jameshcorbett!

jameshcorbett:
Thanks for checking on our standard formats! I would never have thought of that.

PR #299, opened 2025-02-28
dws: storage refactor

jameshcorbett:
Problem: `coral2_dws.py` is too long and complicated and should have
its functionality broken up.

Break out functionality related to monitoring `Storage` resources
into a separate module.

### DIFF
--- a/src/python/flux_k8s/storage.py
+++ b/src/python/flux_k8s/storage.py
@@ -1,0 +1,268 @@
+"""Module defining cleanup routines for the coral2_dws service."""
+
+import logging
+
+import flux
+import flux.kvs
+from flux.hostlist import Hostlist
+from flux_k8s import watch
+from flux_k8s import crd
+
+LOGGER = logging.getLogger(__name__)
+EXCLUDE_PROPERTY = "badrabbit"
+HOSTNAMES_TO_RABBITS = {}  # maps compute hostnames to rabbit names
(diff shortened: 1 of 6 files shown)


src/python/flux_k8s/storage.py
```diff
@@ -0,0 +1,268 @@
...
+
+def _mark_rabbit(handle, status, resource_path, ssdcount, name):
+    """Send RPCs to mark a rabbit as up or down in Fluxion."""
+    if status == "Ready":
+        LOGGER.debug("Marking rabbit %s as up", name)
+        status = "up"
+    else:
+        LOGGER.debug("Marking rabbit %s as down, status is %s", name, status)
+        status = "down"
+    for ssdnum in range(ssdcount):
```
wihobbs, line 215: Are ssds guaranteed to be numbered 1->n? Could `/ssd2/` somehow go away and leave 3 -> n?

wihobbs (approved):
Good job on the comments, btw. They were super helpful for the review!

src/python/flux_k8s/storage.py
jameshcorbett: They are guaranteed to be numbered 1 to N, because the resource graph is generated by a script in this repo, https://github.com/flux-framework/flux-coral2/blob/master/src/cmd/flux-dws2jgf.py which names the ssds sequentially.

jameshcorbett:
Thanks for the review @wihobbs ! Setting MWP.

PR #300, opened 2025-03-04
dws: remove create callback jobtap aux

jameshcorbett:
Similarly to #286, remove a jobtap aux that is set to indicate that a workflow has been created for a job. This aux would cause problems if the plugin were reloaded, so this PR re-architects things so that it is not needed.


cmoussa1 (approved):
LGTM!

PR #302, opened 2025-03-05
dws-jobtap: rpc responses

jameshcorbett:
Problem: the `dws-jobtap` plugin does not respond to the RPCs it
receives. This could lead to the buildup of matchtags in the sender,
in this case `coral2_dws.py`.

Add handling for responding to all RPCs received.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -602,12 +602,12 @@ static void resource_update_msg_cb (flux_t *h,
 {
     flux_plugin_t *p = (flux_plugin_t *)arg;
     json_int_t jobid;
-    json_t *resources = NULL, *errmsg_json, *constraints = NULL;
+    json_t *resources = NULL, *constraints = NULL;
     int copy_offload;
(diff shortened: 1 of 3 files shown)


cmoussa1 (approved):
LGTM!

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -615,31 +615,34 @@ static void resource_update_msg_cb (flux_t *h,
                          "copy-offload",
                          &copy_offload,
                          "errmsg",
-                         &errmsg,
+                         &errmsg_json,
```
grondo, line 618: Would it simplify things to make "errmsg" in the payload an optional string `s?s`, so it doesn't have to be retrieved as a `json_t` and converted to a string later if non-null?

grondo (approved):
LGTM. One suggestion inline.

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: Yeah, I was kind of ambivalent about that but it does make sense. I pushed an extra commit that does that. Setting MWP.

Issue #222, opened 2024-09-26
hang in MPI_Init with unbalanced ranks


ryanday36:
as described in https://rzlc.llnl.gov/jira/browse/ELCAP-705:

(these are all run with `-o mpibind=off`, fwiw)

in a two node allocation (`flux alloc -N2`), running `flux run -n190 ...` puts 96 tasks on one node and 94 on the other and hangs until I ctrl-c.

If I run with `flux run -N2 -n190 ...` flux puts 95 tasks on each node and things run fine (if slowly).

If I use flux's pmi2 (`-o pmi=pmi2` instead of whatever cray mpi is using by default, the original case runs fine.

I did some good old fashioned printf debugging, and it looks like the hang is in MPI_Init, but I haven't gotten any deeper than that. I suspect that this is a an HPE issue, but I'm opening it here too in case you all have any insight. The bit that seems extra confusing is that `flux run -n191 ...` hangs, but `flux run -N2 -n191 ...` doesn't. Both of those should have 96 tasks on one node and 95 on the other, so that doesn't fit super well with my characterization of this as an issue with unbalanced ranks / node.

garlick:
-N might be flipping where the unused core is located.  Example: 2 nodes with 4 cores each
```
ƒ(s=2,d=1) garlick@picl3:~$ flux run -n7 hostname
picl4
picl4
picl4
picl4
picl3
picl3
picl3
ƒ(s=2,d=1) garlick@picl3:~$ flux run -N2 -n7 hostname
picl3
picl3
picl3
picl3
picl4
picl4
picl4
```


It might be worth doing a little audit [here](https://github.com/flux-framework/flux-coral2/blob/master/src/shell/plugins/cray_pals.c#L194-L201) to see if anything stands out with these layouts in mind.

grondo:
I think @garlick meant to put this comment here:
> Have to leave right now but one thing that seems wrong is flux job taskmap shows the same map for both of those cases.
>
> [[0,1,4,1],[1,1,3,1]]
>
> That may be a flux-core bug. Will circle back to this later!

I wonder if the two jobs have the same _R_? I'll try to reproduce this.

garlick:
yes sorry!

grondo:
Hm, this is interesting (did we know this and just forgot?)
```console
$ flux run -N2 -n 7 /bin/true
$ flux job info $(flux job last) R
[...]

Issue #29, opened 2022-09-07
libpals: provide proper `cpus_per_pe`


jameshcorbett:
According to David Gloe at HPE, the [cpus_per_pe member](https://github.com/flux-framework/flux-coral2/blob/63ba22ec96de394588731cf3d15b2d90157f8941/src/shell/plugins/cray_pals.c#L109) of the `pals_cmd_t` struct should be "the number of Linux CPUs (hyperthreads) each PE [i.e. shell task] is bound to." He also noted that

> Overlapping isn't taken into account at all for this. So for example if there are 4 PEs all bound to the same 2 hyperthreads, cpus_per_pe should be set to 2.

The value the flux-coral2 shell plugin provides is not correct because it uses the notion of hwloc cores, not Linux CPUs, and it _does_ account for overlap.

@grondo noted that 

> The distribution of cores to tasks is done in the shell affinity plugin, and as such it is not exported to other plugins at the moment... if you attempt to recreate it, you'd have to use the same code

Thankfully it seems that an improper value (e.g. 0) for the `cpus_per_pe` entry does not cause any errors, and David Gloe guessed that the value isn't read anywhere. So we can punt on the issue for a little while.

grondo:
Just to clarify, by "shell tasks" you mean the user tasks that the job shell launches, not the job shell itself correct?

jameshcorbett:
Yeah, sorry. I was looking for the word and didn't want to say "MPI rank". I should have just said "task" since that's the standard Flux terminology.

PR #303, opened 2025-03-06
add libapinfo convenience library

garlick:
This pulls out the apinfo v1 stuff from the `cray_pals` shell plugin, refactors it, and puts it in a convenience library with unit tests.  I think it'll be pretty easy to add apinfo v5 support as proposed in #282 with a way to fall back to the well tested v1 at runtime.

One thing I did not quite understand - what is the `cpus_per_pe` parameter used for?  I can't for the life of me think why a PMI would need to know how many cpus per task there are, but it'll probably be a "duh" moment when someone tells me.

I did add a test using `libpals` as suggested by HPE although obviously we can't run it in CI at this point.  (It isn't compiled if libpals is unavailable).

Marking WIP pending end to end manual  testing on El Cap.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,6 +98,9 @@ AC_CHECK_FUNCS([gettimeofday localtime_r memset strchr strdup strerror strstr st
 # Checks for packages
 PKG_CHECK_MODULES([SODIUM], [libsodium >= 1.0.14], [], [])
 PKG_CHECK_MODULES([JANSSON], [jansson >= 2.10], [], [])
+PKG_CHECK_MODULES([PALS], [libpals], [HAVE_PALS=yes], [HAVE_PALS=no])
+
+AM_CONDITIONAL([HAVE_PALS], [test "x$HAVE_PALS" = "xyes"])
 
 ##
 # Project directories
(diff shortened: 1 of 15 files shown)


garlick:
Oops, had a missing AM_CPPFLAGS that make an include directive not work with `make dist`.  Also I didn't realize we were clang formatting here.  Fixed those things.

garlick:
I completed some cursory testing on elcap, including double checking for regressions from
- #27
- #222

On the cpus_per_pe thing, I did find this issue which seems to indicate it is not used?  Anyway I don't think this PR changed anything in that area.
- #29

This is probably ready for a review.

grondo (approved):
After a quick review, this LGTM. Should probably get @jameshcorbett to approve as well of course.

jameshcorbett (approved):
LGTM as well thanks!

garlick:
Thanks!  Just fixed a couple of typos, and I'll set MWP and get your work from #282 moved over to this regime.

jameshcorbett:
I'm not sure the unit tests run as part of `make check`... I had to `cd` into `src/common` and run `make check` there. Is that intentional? It doesn't seem desirable.

jameshcorbett:
> I'm not sure the unit tests run as part of `make check`... I had to `cd` into `src/common` and run `make check` there. Is that intentional? It doesn't seem desirable.

Never mind, I just missed the part where the tests run 😊

garlick:
You miss them if you blink :smile:

PR #304, opened 2025-03-06
dws: fix dependency error

### LINKED ISSUE #301: dws-jobtap.c raises job exception if it receives multiple `dws.resource-update` RPCs
On elcap, the `flux-coral2-dws` service was restarted and a bunch of jobs failed with 

```
68519.007s: job.exception ffaN5C7k2Fy type=dws-create severity=0 DWS workflow interactions failed: Failed to remove dependency for job
```

The problem is that, when the `flux-coral2-dws` service restarts, it checks the state of all k8s workflows and sends RPCs to the `dws-jobtap` plugin accordingly. One RPC triggers the jobtap plugin to update a job's jobspec and remove a dependency, and two others cause the jobtap plugin to finish epilog and prolog actions. A restart can cause the `flux-coral2-dws` service to re-send an RPC it had sent before.

The jobtap plugin currently has no way to tell whether it has already received one of the RPCs and taken the corresponding action, and the `flux-coral2-dws` service has no way to tell whether it has already _sent_ one. Multiple dependency-remove RPCs for the same job triggers the jobtap plugin to raise a job exception, which must be avoided.

One solution that comes to mind is to change [this line](https://github.com/flux-framework/flux-coral2/blob/bc160b2368c1ad705054e085a5510feaba541c24/src/job-manager/plugins/dws-jobtap.c#L665) to log an error instead of raising an exception. Not ideal, because there would still be a bunch of spurious error messages logged, but the worst case would be avoided. @grondo, do you have any suggestions?

jameshcorbett:
Fixes #301 .

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -603,8 +603,9 @@ static void resource_update_msg_cb (flux_t *h,
     flux_plugin_t *p = (flux_plugin_t *)arg;
     json_int_t jobid;
     json_t *resources = NULL, *constraints = NULL;
-    int copy_offload;
+    int copy_offload, state;
     const char *errmsg = NULL, *exclude_str;
+    flux_plugin_arg_t *job;
 
     if (flux_msg_unpack (msg,
                          "{s:I, s:o, s:b, s?s, s:s}",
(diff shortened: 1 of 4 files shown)


jameshcorbett:
While working on this PR I found that the issue was trickier than I thought. Issue #301 only occurs if `coral2_dws` is launched with `--disable-fluxion`. Otherwise, when  (upon restart) `coral2_dws` goes to update a jobspec's resources for a second time, it will fail in `directivebreakdown.py` because that module won't accept the jobspec's resources layout. 

This problem doesn't apply if `coral2_dws` is launched with `--disable-fluxion` because in that case the resources are never updated.

grondo (approved):
LGTM!

One suggestion for next time in the tests would be to abstract the launch of the DWS job into a function (that perhaps auto-increments its output file numbers) This would be less likely to trip up someone in the future that adds another restart of the DWS job. (if that is even ever necessary)

jameshcorbett:
Good idea! For now, setting MWP.

PR #309, opened 2025-03-06
dws: fix error logging for invalid directives

jameshcorbett:
Problem: a job with invalid #DW directives currently results in log messages at level ERROR, but really it is a user error and should at most be DEBUG or INFO.

Intercept the problem and raise a UserError instead which will not be logged.


cmoussa1 (approved):
LGTM!

PR #312, opened 2025-03-07
dws2jgf: fix properties on nodes not in DWS

### LINKED ISSUE #311: `dws2jgf` does not apply properties to all nodes
Problem: when `flux-dws2jgf` writes out nodes, it neglects to output properties for all nodes that are not in DWS. For instance, on Tioga, `tioga[42-43]` are not associated with a rabbit, but they are listed in `resource.toml` and are assigned to queue `mi300a`. But `flux dws2jgf` wrote them out without assigning any properties whatsoever.

jameshcorbett:
Fixes #311.


cmoussa1 (approved):
LGTM!

PR #322, opened 2025-03-14
doc: add a rabbit ref, and improve cray-pals man page

garlick:
Problem: we have no docs on the rabbits.

@behlendorf pointed me to a Lustre Users Group presentation that contains detail about the hardware and intended use cases. 

Add this to the References section.

Also, I noticed I missed an option in flux-shell-cray-pals(1), so add that.


jameshcorbett (approved):
👍

Issue #28, opened 2022-09-02
libpals: improve port-distribution mechanism


jameshcorbett:
At the moment the port distribution jobtap plugin is very naive and unoptimized in its distribution of ports to jobs. Every job that runs on two or more nodes is given two unique port numbers. So if the plugin is configured with 1,000 ports, that means there can only be 500 concurrent multi-node jobs.

However, it would be a big improvement if jobs that have disjoint host lists could use the same port.

One possible improvement could be to have the job shells dynamically pick a port.

garlick:
> One possible improvement could be to have the job shells dynamically pick a port.

My first naive thought was maybe the shells could just  use ports derived from the job ID to get a coordinated pseudo-random port for the job without the need for an allocator or inter-shell coordination
```
int my_port  = (jobid % num_ports) + min_port
```

But stupid birthday paradox makes it uncomfortably probable that jobs could fail if there are more than say four sharing a node.

So maybe that plus a coordination handshake, where each shell tries to bind to the candidate port  and reports success or failure back to the leader shell, and keeps retrying with port++ until finally all shells successfully bind and we can continue?  In most cases there would be just one allgather + broadcast so a startup latency comparable to simple PMI?

There is code for the shell allgather + broadcast [here](https://github.com/flux-framework/flux-core/blob/master/src/shell/pmi/pmi_exchange.c) that could be reused.

jameshcorbett:
More details:


>     As an optional enhancement, you could instead use this new method that we've implemented in PALS:
> 
>     Before starting the user's application, find a set of (number of MPMD commands + 1) open ports on all compute nodes in the application by binding sockets to them
>     Pass the port numbers to the application as a comma-separated list in PMI_CONTROL_PORT
>     Pass the socket FDs to the application as a comma-separated list in PMI_CONTROL_FD (make sure not to set FD_CLOEXEC on them)
> 

> If only a single port is provided, my understanding is that if PMI needs additional ports, it will just add one to the port from the environment variable.
> In most cases this works fine, but could potentially cause conflicts if you mix MAPN (multiple apps per node) and MPMD. I haven't seen any bug reports about this causing issues, so it's probably not common.

PR #324, opened 2025-03-14
cray-pals: refactor and add test options

garlick:
Problem: cray-pals isn't set up very well for being extended, and we may need to do some work in it to support VNIs and to improve the port distribution mechanism.

Refactor the plugin to add a global state container that reduces some complexity and provides a way to hold future items that can have their life cycle tied to plugin loading.

Tighten up option parsing.  Add an option to set the pmi port and secret, and then decouple that from the eventlog watching somewhat, so that we can add a collective based method later more easily.

I've played with this a bit and also tested on El Cap with a couple of nodes.  Pretty sure I didn't break anything :-)

### DIFF
--- a/doc/man1/flux-shell-cray-pals.rst
+++ b/doc/man1/flux-shell-cray-pals.rst
@@ -26,18 +26,6 @@ SHELL OPTIONS
   Enable only the cray-pals PMI plugin.  Other PMI implementations may
   be added, separated by commas.
 
-.. option:: cray-pals.no-edit-env
-
-  Prevent the cray-pals PMI plugin from removing Flux's PMI library
-  directory from :envvar:`LD_LIBRARY_PATH`, if present.
-
-  It is removed by default to ensure that Cray PMI library is found
-  before Flux's.
-
-.. option:: cray-pals.apinfo-version=1
-
-  Force HPE apinfo version 1.  The default is version 5.
-
 .. note::
 
   On systems where Cray MPICH is used, it may be helpful for system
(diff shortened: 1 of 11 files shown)


doc/man1/flux-shell-cray-pals.rst
```diff
@@ -26,18 +26,6 @@ SHELL OPTIONS
   Enable only the cray-pals PMI plugin.  Other PMI implementations may
```
jameshcorbett, line 26: typo in commit message: `snall`

jameshcorbett: Also first sentence should have a period?

src/shell/plugins/cray_pals.c
```diff
@@ -271,71 +275,46 @@ static int read_future (flux_future_t *fut,
             return 0;
         }
         if (!strcmp (name, "cray_port_distribution")) {
-            if (json_unpack (context, "{s:o, s:I}", "ports", &array, "random_integer", random)
+            if (json_unpack (context,
+                             "{s:[ii] s:I}",
```
jameshcorbett, line 279: I _think_ there was a thought that in the future the plugin might not always send or receive exactly two integers? 

Looked back in my email history and I found:

> As an optional enhancement, you could instead use this new method that we've implemented in PALS:
1. Before starting the user's application, find a set of (number of MPMD commands + 1) open ports on all compute nodes in the application by binding sockets to them
2. Pass the port numbers to the application as a comma-separated list in PMI_CONTROL_PORT
3. Pass the socket FDs to the application as a comma-separated list in PMI_CONTROL_FD (make sure not to set FD_CLOEXEC on them)

So I guess it's only if Flux were to support MPMD commands I guess.

jameshcorbett (approved):
Seems reasonable

src/shell/plugins/cray_pals.c
garlick: Binding to the ports ahead of time and passing file descriptors makes a lot of sense!  We should do that  as part of improving the port allocation method (#28)

As for MPMD support, hmm.  I'm not sure if/when we'll need it but it still would be good to not close off that aspect of the design.  Perhaps I could add a note that reflects what you've written here so we can keep that in mind going forward

doc/man1/flux-shell-cray-pals.rst
garlick: Wow, almost as bad as misspelling "typo" :-)

src/shell/plugins/cray_pals.c
garlick: Oh I missed it but you already had commented in the code
> * The number of ports that must be specified is (# of MPMD commands) + 1.

I'll add something in the struct I added also.

src/shell/plugins/cray_pals.c
```diff
@@ -73,11 +73,19 @@
  * the PMI_CONTROL_PORT distribution mechanism.
  */
 
+struct pmi_bootstrap_info {
+    int port[2];    // in the future, make array size dyanmic for MPMD
```
jameshcorbett, line 77: typo "dynamic" not even worth a comment but it looks like you'll have to rebase for formatting anyway

garlick: Doh!  Sorry about that.  OK fixed, and I'll set MWP. 

Thanks!

PR #327, opened 2025-03-18
dws: fix add-constraint logic

### LINKED ISSUE #276: dws-jobtap.c: fix constraints
Problem: `dws-jobtap.c` [tries to add a `-property:badrabbit` constraint into users' jobspecs](https://github.com/flux-framework/flux-coral2/blob/8abdb50447589c20985777da9347ffb2db0891ba/src/job-manager/plugins/dws-jobtap.c#L582). However it does not do so correctly and generates invalid jobspecs.

There must be some logic in `flux-core` for adding constraints after submission time, because it is done for queues. Perhaps that could be copied over. @grondo do you have a pointer by chance?

jameshcorbett:
Fixes #276 .


cmoussa1 (approved):
LGTM!

PR #335, opened 2025-04-01
ci: add coverage action

jameshcorbett:
Problem: there is no coverage testing for this repo.

Add coverage testing.


garlick (approved):
I'm a bit clueless about how this stuff works, however I don't see a codecov report posted in CI?

I'll tentatively approve but we might want to ask @grondo to have a peek here.

PR #336, opened 2025-04-03
dws: epilog timeout

### LINKED ISSUE #321: Add configurable timeout for dws-epilog action
Sometimes workflows can get stuck in `Teardown`, see e.g. #313. This aggravates admins and users because the jobs can be occupying huge numbers of nodes.

After discussion today, it seems like the path forward is the following:

1. Add a special exception name/type for abandoning workflows and freeing resources
2. Add a timeout in the `dws-jobtap` plugin after starting the `dws-epilog` action
3. If the timeout occurs, raise the special exception
4. In the exception event callback, if the special exception is found, remove the epilog event and send an RPC
5. The RPC recipient should look at the `servers` resource for the associated workflow and see which rabbits have a `capacity` > 0, since that indicates they have not fully cleaned up. Those rabbits should be disabled with an annotation indicating why.
6. The RPC recipient should also look at the `clientmount` resources for the associated workflow and drain any nodes that have not unmounted their file systems.

jameshcorbett:
Fixes #321.

This is a big PR and some of the python changes still need tests, but I have to figure out how to write them because they'll need to be able to inspect kubernetes and there is an opportunity for all kinds of race conditions.

@grondo do you think you could look at just the first three commits? They deal with the jobtap plugin.


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -30,12 +30,15 @@
...
 #define DWS_EPILOG_NAME "dws-epilog"
+#define EPILOG_ABORT_EXCEPTION "dws-epilog-timeout"
 
 struct create_arg_t {
     flux_plugin_t *p;
     flux_jobid_t id;
 };
 
+static double EPILOG_TIMEOUT = 0.0;
+
```
grondo, line 41: Since this variable can be updated, I'd use lower case to differentiate from a constant (which are typically all upper case)

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -776,6 +829,16 @@ static const struct flux_plugin_handler tab[] = {
...
+    flux_t *h = flux_jobtap_get_flux (p);
+
+    if (flux_plugin_conf_unpack (p, "{s?f}", "epilog-timeout", &EPILOG_TIMEOUT) < 0 && h) {
+        flux_log (h,
+                  LOG_NOTICE,
+                  PLUGIN_NAME ": failed to unpack config: %s",
+                  flux_plugin_strerror (p));
+    }
+    if (EPILOG_TIMEOUT <= 0 && h)
+        flux_log (h, LOG_NOTICE, PLUGIN_NAME ": not configured with epilog timeout");
```
grondo, line 864: would it be useful to have the epilog timeout logged if configured too? i.e. just always execute
```c
flux_log (h, LOG_INFO, PLUGIN_NAME ": epilog timeout = %.3fs", epilog_timeout);
```
(fine as is, I'm just not used to seeing log messages about something "not configured")

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -460,6 +494,25 @@ static int cleanup_cb (flux_plugin_t *p, const char *topic, flux_plugin_arg_t *a
...
+                                            watcher,
+                                            (flux_free_f)flux_watcher_destroy)
+                       < 0) {
+                dws_epilog_finish (h, p, id, 0, "Failed to init " DWS_EPILOG_NAME " timeout");
+                flux_log_error (h, "Failed to init " DWS_EPILOG_NAME " timeout for %s", idf58 (id));
+                flux_watcher_destroy (watcher);
+                return -1;
+            }
+            flux_watcher_start (watcher);
+        }
```
grondo, line 515: If the plugin is unloaded while the watcher is still active, this could cause a broker crash since the `epilog_timeout_cb` symbol won't exist anymore.

Unfortunately, in these situations, the plugin needs to keep a list of outstanding watchers so it can stop them and/or destroy them on exit. i.e. you can't rely on using the job aux item for this purpose.

src/job-manager/plugins/dws-jobtap.c
```diff
@@ -577,12 +580,32 @@ static int exception_cb (flux_plugin_t *p, const char *topic, flux_plugin_arg_t
...
-                  flux_rpc_pack (h, "dws.teardown", FLUX_NODEID_ANY, 0, "{s:I}", "jobid", id))) {
-            flux_log_error (h, "Failed to send dws.teardown RPC for job %s", idf58 (id));
-            return -1;
+        if (type && (strcmp (type, EPILOG_ABORT_EXCEPTION) == 0)) {
+            if (!(teardown_fut =
+                      flux_rpc_pack (h, "dws.abort", FLUX_NODEID_ANY, 0, "{s:I}", "jobid", id))) {
+                flux_log_error (h, "Failed to send dws.abort RPC for job %s", idf58 (id));
+                return -1;
+            }
+            flux_future_destroy (teardown_fut);
```
grondo, line 589: For RPCs that do not have a response, consider adding `FLUX_RPC_NORESPONSE` so that a matchtag doesn't have to be allocated:
```
       FLUX_RPC_NORESPONSE
              No response is expected. The request  will  not  be  assigned  a
              matchtag,  and  the  returned  flux_future_t is immediately ful‐
              filled, and may simply be destroyed.
```
Same below I assume.

jameshcorbett: It's not an error if a response is given to a `FLUX_RPC_NORESPONSE` RPC though is it? Because a response is given, although it's currently ignored.

grondo: Yeah, there'd be an error I think, so nevermind. I'd add a comment then that the expected response is ignored. (sorry for not understanding the code)

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: The same is true for `flux_future_t`s with callbacks that are functions defined in the plugin as well right? If so then it's a wider problem and I might try to address it all at once in a later PR.

How could I trigger cleanup on plugin removal? By using `flux_plugin_set_aux` or something, with the destructor set to my cleanup function?

grondo: > The same is true for flux_future_ts with callbacks that are functions defined in the plugin as well right? If so then it's a wider problem and I might try to address it all at once in a later PR.

Yes, that is fine to leave for a future commit. I just happened to notice it here.

> How could I trigger cleanup on plugin removal? By using flux_plugin_set_aux or something, with the destructor set to my cleanup function?

Yeah, I think the approach would be to have a global context stored in the plugin aux cache and add a list of objects to that ctx that would need to be freed on plugin removal.

Actually looking at existing plugins in core, many of them suffer from this same issue :facepalm: . Feel free to add an issue to address this at some point in future. For now we'll have to be careful reloading most plugins that do asynchronous work.

src/modules/coral2_dws.py
```diff
@@ -306,12 +308,98 @@ def abort_cb(handle, _t, msg, k8s_api):
...
+        except (KeyError, TypeError) as exc:
+            LOGGER.warning(
+                "error inspecting mounts for %s resource found for workflow %s: %s",
+                crd.CLIENTMOUNT_CRD.plural,
+                workflow_name,
+                exc,
+            )
+            # assume node bad
+            to_drain.append(node)
+            continue
```
cmoussa1, line 425: Does this last `except` block need a `continue` statement here? Or will it already just iterate to the next `resource` in `clientmounts`?

cmoussa1 (commented):
A couple of small clarifying questions after a pass but I think most of the Python stuff looks good to me so far!

src/job-manager/plugins/dws-jobtap.c
jameshcorbett: I opened issue #337 for this

t/t1000-dws-dependencies.t
```diff
@@ -224,4 +231,12 @@ test_expect_success 'job-manager: dws jobtap plugin handles --requires=^property
 	flux job wait-event -vt 5 ${create_jobid} clean
 '
 
+test_expect_success 'job-manager: load dws-jobtap plugin with epilog timeout > 0' '
+	flux jobtap remove dws-jobtap.so &&
+	flux dmesg --clear &&
+	flux jobtap load ${PLUGINPATH}/dws-jobtap.so epilog-timeout=0.1 &&
+	flux dmesg | test_must_fail grep "dws: not configured with epilog timeout" &&
+	flux dmesg | test_must_fail grep "dws: failed to unpack config"
```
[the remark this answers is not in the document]
jameshcorbett: The tests are expected to fail because the loge message should not be present, because the config should have been unpacked successfully and the timeout should be configured. Going off one of Mark's comments though I changed the log message.

src/modules/coral2_dws.py
jameshcorbett: Good catch, removed!

jameshcorbett:
Ok I think I've addressed all the comments so far, or opened issues for the future. @cmoussa1 I added a test for the new Python stuff. Other than addressing the comments and changing the name of a variable or two, it should be exactly the same as before except for the new test in the last commit.

cmoussa1 (approved):
this LGTM!

PR #345, opened 2025-04-14
dws: fix missing argument to `log_rpc_response`

jameshcorbett:
Problem: a future callback invoking `log_rpc_response` is missing an argument.

Fix it.


cmoussa1 (approved):
LGTM!

PR #346, opened 2025-04-17
dws: catch remove-finalizer exception and retry

jameshcorbett:
Problem: on Hetchy, Flux failed to remove finalizers from a workflow due to a race condition. The workflow had already been deleted, perhaps manually, and attempts to remove the finalizer failed because the `remove_finalizer` function was using an old version of the workflow with finalizers that had already been removed. Therefore, the calls to patch the workflow to remove the Flux finalizer were seen by kubernetes as an attempt to *add* finalizers, which is not allowed if the workflow has been deleted, and a 422 error code was returned.

Rename the old `remove_finalizer` to `_remove_finalizer` and add a new exception-catching function `remove_finalizer` that wraps `_remove_finalizer`, fetching the latest workflow and trying again.

### DIFF
--- a/src/python/flux_k8s/cleanup.py
+++ b/src/python/flux_k8s/cleanup.py
@@ -16,6 +16,26 @@ CLEANUP_LOOP = asyncio.get_event_loop()
 
 
 def remove_finalizer(workflow_name, k8s_api, workflow):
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM!

Issue #256, opened 2025-01-28
Support for secure data movement


jameshcorbett:
In the rabbit meeting today we discussed a need for Flux to grab secrets from Kubernetes and export them as environment variables to Flux jobs. As far as I understand it, Flux will need to look for a flag in a workflow (or maybe a directivebreakdown) and then, if the flag is set, grab a secret from Kubernetes and add it as an environment variable.

@grondo this will I think require that the eventlog be secure, so one user can't look at the events for another user's job, because currently flux-coral2 puts the environment variables in the event log for the shell to fetch and set. We already guarantee that security though right? 

I remember some months ago there was a discussion (with Kalan I think) about the safety of sharing event logs, but maybe that was just informally, like between us.

grondo:
> this will I think require that the eventlog be secure, so one user can't look at the events for another user's job,

Yes, fetching the eventlog as a guest user requires accessing the KVS through the job-info service, which restricts users to data only within their own jobs. E.g. if I try to look at the eventlog of another user's job:

```console
$ flux job eventlog -H f445SZmdRv7q
flux-job: flux_job_eventlog_lookup_get: Operation not permitted
```

Neither can I access the KVS directory of the job directly:

```console
$ flux kvs dir $(flux job id --to=kvs f445SZmdRv7q)
flux-kvs: job.1240.c1e8.b100.0800: Operation not permitted
```
However, users with `sudo` access to `flux` or `root` would be able to fetch these eventlogs since they can operate as the instance owner.

grondo:
> I remember some months ago there was a discussion (with Kalan I think) about the safety of sharing event logs, but maybe that was just informally, like between us.

We do tend to heedlessly copy and paste eventlogs into issues, slack and MM messages since there is not much if anything security significant in there now. I wonder if there's some way we can automatically obscure sensitive information in an eventlog context, kind of like github workflows do in output. 

Are these secrets going to be the same for every job, or will a new secret be created per job, reducing the impact of a potential compromise?

roehrich-hpe:
> Are these secrets going to be the same for every job, or will a new secret be created per job, reducing the impact of a potential compromise?

The secret will be unique per Workflow. James, is there a 1-1 relationship between Workflow and job?

jameshcorbett:
> > Are these secrets going to be the same for every job, or will a new secret be created per job, reducing the impact of a potential compromise?
> 
[...]

PR #347, opened 2025-04-23
dws: secret token for copy-offload

jameshcorbett:
Problem: in order to support the new style of rabbit data movement,
Flux needs to grab a kubernetes secret for certain workflows, and
then export it to jobs as an environment variable.

Grab the secret and pack it into the prolog-remove RPC for it to
be set in the job's environment by the job shell.

Fixes https://github.com/flux-framework/flux-coral2/issues/256.

### DIFF
--- a/src/modules/coral2_dws.py
+++ b/src/modules/coral2_dws.py
@@ -19,7 +19,10 @@ import time
 import re
 import contextlib
 from datetime import datetime
+import base64
 
+import kubernetes
+import kubernetes.client
 from kubernetes.client.rest import ApiException
 import urllib3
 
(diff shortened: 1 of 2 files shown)


src/modules/coral2_dws.py
```diff
@@ -122,6 +125,22 @@ def owner_uid(handle):
...
+def fetch_job_environment(secrets_api, workflow):
+    """Fetch the variables to be exported to a user's job."""
+    variables = workflow["status"].get("env", {})
+    token = workflow["status"].get("workflowToken")
+    if token is None:
+        return variables
+    # if the workflow requires a secret, fetch it
+    secret = secrets_api.read_namespaced_secret(
+        token["secretName"], token["secretNamespace"]
+    )
```
cmoussa1, line 137: Forgive my limited understanding of kubernetes, but does not every workflow need a `secret`? If so, does this API call to `.read_namespaced_secret()` need to be prefaced by some `if` condition or something? Or is this okay to call this every time? Totally fine to leave this as is if so.

cmoussa1 (approved):
LGTM! One clarifying question because I don't fully understand Kubernetes 😅

src/modules/coral2_dws.py
jameshcorbett: Good question! There's an early `return` just above this line:

```
    token = workflow["status"].get("workflowToken")
    if token is None:
        return variables
```

If the workflow has a `.status.workflowToken` field then it's expected to have a secret and Flux is expected to export it as an environment variable. Otherwise the `return` will skip fetching the secret (which would fail probably).

cmoussa1: Great! Thanks for the clarification 👍

roehrich-hpe: Correct. Check every workflow for status.workflowToken, but allow that the field may be empty. Most workflows won't have one.

PR #351, opened 2025-04-28
dws: update dws systemstatus resource

### LINKED ISSUE #341: dws: handling node crashes
Problem: currently when nodes crash in a rabbit job, the rabbit job can become stuck because DWS cannot tell what the status of the mounted file systems are. 

After discussion with @behlendorf, it seems that since Flux has some information about node status, it should be able to help out. If the system instance could monitor node crashes in rabbit jobs, it could drain those nodes and modify a kubernetes resource (specifically, remove a finalizer from a `clientmount` resource) so the job can clean up properly. 

Proposed implementation:

Node failures are posted to a job's eventlog as an exception of severity 2. The `dws-jobtap` plugin could monitor for those exceptions and then send two RPCs:

1. A `resource.drain` RPC, to drain the affected node(s) or, if already drained, to update the drain reason to indicate that a rabbit file system needs to be unmounted, so that admins know that some additional steps may need to be taken.
2. A new custom RPC, telling a service to remove a finalizer from a kubernetes resource.

jameshcorbett:
As described in #341, DWS needs Flux to tell it what nodes are up or down so it can potentially forget about some rabbit file system mounts, allowing jobs to clean up.

Fixes #341 .

### DIFF
--- a/doc/guide/rabbit.rst
+++ b/doc/guide/rabbit.rst
@@ -145,9 +145,16 @@ in a batch script:
 Fetching Rabbit Information
 ---------------------------
 
-:man1:`flux-getrabbit` can be used to look up the rabbits used by a job,
-as well as what rabbits have PCIe links to which compute nodes and vice
+The :man1:`flux-getrabbit` command can be used to look up the rabbits used by
+a job, as well as what rabbits have PCIe links to which compute nodes and vice
 versa.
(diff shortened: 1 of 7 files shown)


cmoussa1 (approved):
LGTM! nice work

PR #353, opened 2025-04-30
dws: fix memory leak in callback

jameshcorbett:
Problem: in `cleanup_cb`, there are multiple `return` statements after some memory has been allocated but before the memory is set to a jobtap aux for automatic deletion. Those `return` statements would therefore cause the memory to be leaked.

Assign the memory to a jobtap aux at the time of allocation.


cmoussa1 (approved):
LGTM!

PR #354, opened 2025-05-01
dws: teardown timeout

jameshcorbett:
Problem: PR https://github.com/flux-framework/flux-coral2/pull/336 added an epilog timeout for moving a workflow to
teardown, as well as removing the epilog and potentially draining
nodes and disabling rabbits. However, before taking that drastic step
it makes sense to move a workflow to Teardown to give it a chance
to clean up. In theory, this could make it so the epilog timeout
never hits or at least reduce the number of rabbits disabled when
the timeout hits.

Add a timer-watcher that moves a workflow to Teardown.


cmoussa1 (approved):
LGTM!

PR #358, opened 2025-05-12
dws: postrun timeout

jameshcorbett:
There have recently been several issues on elcap clusters where rabbit workflows were stuck in the `PostRun` state waiting for nodes to unmount file systems. Some of them had a hung mount point. Flux needs a way to abandon mounts and drain nodes when this happens, and this PR implements that.

### DIFF
--- a/src/python/flux_k8s/systemstatus.py
+++ b/src/python/flux_k8s/systemstatus.py
@@ -1,6 +1,8 @@
 """Module defining routines for handling k8s SystemStatus resources."""
 
+import itertools
 import logging
+import time
 
 from kubernetes.client.rest import ApiException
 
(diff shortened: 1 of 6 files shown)


src/python/flux_k8s/systemstatus.py
```diff
@@ -10,33 +10,59 @@
...
+        """Patch the systemstatus kubernetes object with the latest disabled nodes.
+
+        A disabled node is one that either has an offline broker or a failed unmount.
+        """
+        nodes = {node_name: "Enabled" for node_name in newly_online}
+        for node_name in newly_offline:
+            nodes[node_name] = "Disabled"
+        self.k8s_api.patch_namespaced_custom_object(
+            *crd.SYSTEMSTATUS_CRD, "default", {"data": {"nodes": nodes}}
+        )
```
cmoussa1, line 99: commit message nit: "will needs to" --> "will need to"

t/t1002-dws-workflow-obj.t
```diff
@@ -784,13 +784,70 @@ test_expect_success 'job submission with valid DW string works with teardown_aft
...
+		${jobid} epilog-finish &&
+	flux job wait-event -vt 15 ${jobid} clean
+'
+
+test_expect_success 'systemstatus object is updated' '
+	kubectl get systemstatus default -ojson | \
+		jq -e ".data.nodes.\"$(hostname)\" == \"Disabled\"" &&
+	flux resource undrain $(hostname) &&
+	date &&
+	sleep 10 &&
```
cmoussa1, line 834: just double checking that this `sleep 10` here is intended 👍

jameshcorbett: It's not! Good catch. It should sleep for a second or two, but not 10.

src/python/flux_k8s/systemstatus.py
jameshcorbett: Thanks!

jameshcorbett:
Force-pushed to fix some issues, setting MWP.

PR #359, opened 2025-05-13
dws: heartbeat and profile data

jameshcorbett:
Problem: sometimes the coral2_dws script becomes seemingly
unresponsive. It would be useful if it printed out a heartbeat
message and some profiling data.

Add a heartbeat message and profiling data.


cmoussa1 (approved):
Looks really neat!

PR #360, opened 2025-05-16
dws: status rpc

jameshcorbett:
Problem: sometimes the coral2-dws service becomes unresponsive. It
would be helpful if there were a RPC that could check the status of
the service.

Add a dws.status message handler.


cmoussa1 (approved):
Awesome! Looks good

jameshcorbett:
Thanks! Setting MWP.

PR #362, opened 2025-05-22
doc: fix link and style

jameshcorbett:
Problem: the docs have a broken link and some incorrect formatting, putting some words in single backticks rather than double.

Fix them.


cmoussa1 (approved):
Nice!

PR #363, opened 2025-05-22
dws: remove python 3.7 feature

jameshcorbett:
Problem: the pstats SortKey enum was introduced in Python 3.7, but code must support 3.6.

Use a string instead of the SortKey enum.


cmoussa1 (approved):
LGTM!

PR #366, opened 2025-06-04
cmd: fix help messages

jameshcorbett:
Problem: due to the ordering of arguments, the help message for `flux getrabbit` and other commands displays text like "--computes, -c=HOSTS" which looks strange.

Change the order to make it print text like "-c, --computes=HOSTS".


cmoussa1 (approved):
Nice improvement!

Issue #344, opened 2025-04-09
implement hardware collectives for slingshot


garlick:
Let's collect notes on what's needed to support hardware collectives on El Cap.

I'll summarize what I got from talking to @trws just now -  I may be misremembering so please correct me.

HPE arranged a test recently and demonstrated a clear benefit  of hardware collectives on job sizes above about 4000 nodes.  Below that not so much.  Since the resources required to enable collectives are somewhat limited, it may be best to *only* enable them on large jobs, and even then, perhaps only when requested.  Performance can actually degrade when collectives are used on smaller jobs.

The process involves making one request to the fabric  manager per job to set up (up to 24?) "trees".   A handle is returned.  The handle must be distributed to all the MPI tasks via environment variables.  (The actual consumer is libfabric, which is used by Cray MPI).  At the end of the job, the fabric manager is told to destroy the handle.  The fabric manager operations could be performed by the the job shell, authenticated as the end user.  Thus a job shell plugin could handle making these requests and distributing the handle to other parts of the job, and cleaning up afterwards.  All self contained within the job. 

The details of the authentication need to be understood.  There were two options, a pre shared key, which would presumably implies that Flux would need to do key management for users; and an Oauth like mechanism.

bronis:
I do not seem to have the HPE presentation with the graph of the results from the demonstration runs handy but my recollection is that the benefit starts being significant at much smaller node counts than 4,000. I am fairly certain it was significant at node counts > 1,000 (at most).

bronis:
The requirement was that HPE would provide a mechanism to enable HW collectives for jobs running on 25% or more of the system nodes (meaning the lower bound depends on system size so many fewer nodes on Tuolumne) for MPI_COMM_WORLD and copies of MPI_COMM_WORLD. I believe the actual implementation is a bit more flexible than that but the results indicated that was a reasonable limitation to impose.

With respect to my previous comment, I know that the results indicated they would be beneficial at least for full system jobs on Tuolumne and I am fairly certain even the 25% (so about 280 nodes) level was significant.

Again on the performance front, I do not recall seeing any evidence of performance degradations. The issue is that smaller jobs are not guaranteed to have the switch resources -- the point of the minimum size requirement is to avoid having to solve an NP-hard problem in setting up the trees, not to ensure jobs are large enough to perform well.

garlick:
Thanks Bronis, understood. I'm sure we can easily make these thresholds site-configurable.

[...]

Issue #24, opened 2022-08-23
MPI: Integrate with HPE's CXI library for allocating VNIs


jameshcorbett:
> Per the latest batch of emails with Cray, it looks like the Shasta APIs can be used a la carte by the WLM.
> 
> APIs that seem like ones that we would use:
> 
>     HMS's Hardware Inventory API (combined with the ClusterStor Inventory API)
>     The ATOM node health check API
> 
> Based on the latest emails and discussions, we will not be using either of these interfaces.  We will only be using the node health checks through the default TOSS4 utility (nodediag?).
> 
> CXI library for allocating/setting up VNIs - requires root

jameshcorbett:
Slingshot requires the use of VNIs (think of VLANs).  If you use the same VNI for everything, eventually you exhaust endpoints on the switches.  Slurm will be using one VNI per job step.

For Flux:

    Day 0: pre-allocate X VNIs per sub-instance, then launches within that sub-instance round-robin across VNIs
    Day 1: allow users to request extra VNIs per sub-instance
    Day 2: bespoke setuid binary that when run at top-level will do anything but when run as a user, it is limited to what the top-level constrained things to

garlick:
See also:  https://github.com/SchedMD/slurm/tree/master/src/plugins/switch/hpe_slingshot

garlick:
VNI tagging was brought up again recently in a (not public) TOSS issue:   https://lc.llnl.gov/jira/browse/TOSS-5932

This statement from the issue seemed like a good description of the problem:

> As part of the changes that implement VNI tagging on the HPE Slingshot NIC, as of Slingshot 2.0.1 the default CXI Service has been disabled. This means that deployments must implement additional host-side configuration (using the job scheduler plug-ins for example) implement VNI tagging, or explicitly re-enable the default service to have applications operate as in previous releases. (This also means that CXI diagnostics need to pass the VNI information on the command line). HPE recommends fully implementing VNI tagging for isolating RDMA traffic to protect against memory writes from nodes not known to be part of the job. Refer to Section 8.3. of the HPE Slingshot Operations Guide - Customer for more information.

trws:
I don't see the context elsewhere, or another issue, so I'll add it here.  We need to implement VNI assignment at least local to each node.  The switch reconfiguration, which is the part that has performance and interface concerns, we don't need to deal with, but it's also possible to exhaust resources _on the NIC_ if we don't.  From what I understand, there are two parts to this.

[...]

PR #369, opened 2025-06-10
add design document for slingshot VNI + hw collectives

garlick:
This is a first cut at a design document for Flux slingshot support.  This would enable
- #24
- #344 
- NIC resource management
- traffic class restriction

This is marked as a WIP pending
- some prototyping to make sure this is going to work
- filling out some missing details in the fabric manager portion

Despite being a WIP, feedback would be very much appreciated!


jameshcorbett (commented):
First three commits look reasonable to me, I see the fourth is far larger though, I'll have to circle back to it later!

garlick:
Yeah sorry about that. It's kind of a lot.

garlick:
In the interest of keeping things moving forward, maybe we could just get this document merged as is and revise it as we learn things while prototyping?  I'll drop the WIP just in case that's agreeable.

We could also leave it pending until others have had a chance to weigh in.  Up to you @jameshcorbett

doc/guide/slingshot.rst
```diff
@@ -0,0 +1,594 @@
...
+The above components are extended and a new component is added:
+
+rest-client broker module
+  An general REST client module based on libcurl that can be loaded
+  under different names to provide a Flux RPC gateway to a REST service.
+  It is loaded as cray-slingshot-fabric on the leader broker and is used
+  as a gateway to the Slingshot fabric manager.
+
+  It uses the libcurl "multi" interface [#curlmulti]_ to remain responsive
+  while waiting for REST servers to respond.
```
grondo, line 564: Just thinking out loud here, but the REST client may be easier to implement as a Python systemd service that does the RPC to REST gateway (similar to the services already provided in flux-coral2 and flux-accounting)

doc/guide/slingshot.rst
```diff
@@ -0,0 +1,594 @@
...
+Within the sub-instance the following occurs:
+
+#. At startup, the broker module on each rank reads the Slingshot environment
+   variables from the broker environment and creates an inherited reservation
+   that will be re-used for all jobs.
+
+#. The job shell plugin asks the local broker module for the job's VNI
+   reservation and local CXI service IDs, then launches the job with the
+   appropriate environment variables set.
+
```
grondo, line 545: I'm going to ask a dumb question here: Will the environment variables set for the job actually be any different than if they were just passed through from the parent environment? (i.e. skip the broker module and shell plugin in subinstances?)

I guess I'm wondering if the environment variables are going to be the same for all jobs if there would be some way to skip the job shell -> module communication here.

grondo (approved):
Nice! I learned a lot reading this through. Seems fine to merge if we get an ACK from @jameshcorbett. I just had a couple questions on a quick skim.

doc/guide/slingshot.rst
garlick: The environment for all jobs is the same, but the environment on each node of the job is different.   The job environment is initialized from the environment of the initial program, which comes from the rank 0 broker.

The reasons it is different on each node is
- SLINGSHOT_DEVICES reflects the Cassini devices on the local node, but some nodes might have a different number of them, or one might be unusable but we want jobs to run anyway (detecting bad hardware is not in scope here)
- SLINGSHOT_SVC_IDS reflects the CXI service allocated on each Cassini device.  The same VNI goes in to each CXI service allocation request, but the driver hands back a service ID that could be different on each node/interface.

There's no off-node messaging in the transfer of the local environment variables from the broker to the shell, and since the code needs to be there anyway for the system instance case, it didn't seem like a bad solution?

doc/guide/slingshot.rst
grondo: Thanks for the clarification! In my quick skim I didn't catch that this service is required on subinstances as well.

Even so, it may be worth seeing if we could spawn a Python service from rc1 or something as an alternative. I could be wrong, but writing the REST code in Python maybe be a big enough time saver to warrant some extra work in making Python-process-as-a-service workable (I mean, it would work now, but we just don't have a good way to track those services)

Anyway, sorry for the diversion. That discussion can obviously come later. I was more wondering why libcurl was specified here, but maybe I was reading this as an RFC SHALL and it is more a description of a possible implementation.

garlick: Great point and not a diversion!

doc/guide/slingshot.rst
garlick: Yeah, that potential gotcha is worth considering.  There will be a way to turn the whole thing off with e.g. `-o cray-slingshot=off`, and potentially we could do that for single node jobs by default.  Maybe there are other ways to think about it too.

jameshcorbett (approved):
I'm good with this!

garlick:
Cool, thanks.  I'll set MWP.

PR #370, opened 2025-06-24
simplify and rename port allocator

garlick:
Problem: the port allocator fetches R from the KVS, but it's now available as a plugin IN argument.

Drop the KVS fetch and associated prolog synchronization, since the event containing the ports can be reliably posted before the start event now.

Also, rename the jobtap plugin to `cray-pmi-bootstrap`.  The  old name of `cray_pals_port_allocator` is outdated, since the plugin now sets the PMI shared secret too, and the ports are just for PMI not for PALS.

Tack on a couple of other minor cleanups.

We should double check that we aren't configuring the plugin under the old name anywhere so I'll mark this WIP pending that.

### DIFF
--- a/doc/guide/admin.rst
+++ b/doc/guide/admin.rst
@@ -73,7 +73,7 @@ in all Flux instances.  Edit ``/etc/flux/shell/initrc.lua`` to contain:
       shell.options['pmi'] = 'cray-pals,simple'
   end
 
-The ``cray_pals_port_distributor.so`` jobtap plugin, required by the above,
+The ``cray-pmi-bootstrap.so`` jobtap plugin, required by the above,
 is loaded automatically via ``/etc/flux/rc1.d/01-coral2-rc``.
 
 *******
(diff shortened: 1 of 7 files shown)


garlick:
> We should double check that we aren't configuring the plugin under the old name anywhere so I'll mark this WIP pending that.

Yes we are.  From tuo:
```toml
[job-manager]
plugins = [
  { load = "perilog.so" },
  { load = "cray_pals_port_distributor.so", conf = { port-min = 11000, port-max = 12000 } },
  { load = "dws-jobtap.so", conf = { epilog-timeout = 5400.0 } }
]
```
however, this has no effect.  The plugin is loaded by `rc1.d/01-coral-rc` in all instances, and the configured values are the same as the defaults.  We could ask the admins to remove this at any time.

jameshcorbett (approved):
Code changes look great, and I opened an Ansible PR to remove the jobtap plugin from the config. https://izlc.llnl.gov/gitlab/LC/ansible/-/merge_requests/19564

garlick:
Sorry, I spotted some newly dead code and force pushed that change too.  Meanwhile, it seems the noble test is failing to download kubectl or something.  Maybe a transient problem, so I'll just come back in a little while and rerun that test.

jameshcorbett (approved):
Looked through again, still looks good! Test is passing now, I'd say set MWP whenever you're ready.

PR #371, opened 2025-06-26
dws: version update

jameshcorbett:
Problem: the dws and nnf CRD versions are old.

Update them, and fix some other miscellaneous test problems.


cmoussa1 (approved):
LGTM!

PR #377, opened 2025-07-02
docs: add rabbit references

jameshcorbett:
Problem: there should be more references to the HPE rabbit
documentation in this repo's rabbit docs, because the HPE docs have
a lot of additional detail. Currently there is only one link.

Add more links.


garlick (approved):
Nice!

PR #378, opened 2025-07-24
dws: fix typo in systemd unit file

jameshcorbett:
Problem: the systemd watchdog introduced in PR #374 has a typo.

Fix it.

See https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#WatchdogSec=


cmoussa1 (approved):
LGTM!

PR #381, opened 2025-08-01
dws: exceptions raised while fetching pod logs are lost

jameshcorbett:
Problem: the `save_pod_log` function of the `flux_k8s.cleanup` module
sometimes raises exceptions, but the tracebacks are lost.

Log exceptions from the function.


cmoussa1 (approved):
LGTM!

PR #383, opened 2025-08-06
 Open add support for modprobe based startup and shutdown

garlick:
Prepare for flux-core modprobe based startup.

This mostly mimics @grondo's work in flux-framework/flux-sched#1380 including some commit message cut and paste.  Hope that's OK.


garlick:
NP if we want to hold off.  I guess I am enthusiastic about moving fowrard. :-)

etc/modprobe/rc1.d/coral2.py
```diff
@@ -0,0 +1,28 @@
...
+    requires=["job-manager"],
+)
+def coral2_jobtaps_all(context):
+    context.bash(
+        """
+    for mod in cray-pmi-bootstrap.so; do
+        flux jobtap query $mod >/dev/null 2>&1 || flux jobtap load $mod
+    done
+    """
+    )
```
grondo, line 28: You could actually do this in Python more efficiently as (not tested though)
```python
def coral2_jobtaps_all(context):
    for mod in ("cray-pmi-bootstrap"):
        try:
            context.rpc("job-manager.jobtap", {"load": mod}).get()
        except FileExistsError:
            pass
```

etc/Makefile.am
```diff
@@ -1,3 +1,8 @@
 #if HAVE_SYSTEMD
 systemdsystemunit_DATA = flux-coral2-dws.service
 #endif
+
+fluxconfdir = $(sysconfdir)/flux
+
+nobase_dist_fluxconf_SCRIPTS = \
+	modprobe/rc1.d/coral2.py
```
grondo, line 8: commit message nit:
The problem statement is not quite correct. The existing setup here should work fine with modprobe. So maybe just "doesn't take advantage of new modprobe features"

grondo:
Already approved and LGTM, so feel free to set MWP as far as I'm concerned!

jameshcorbett:
I'd like to read through the core and sched modprobe PRs first so I can understand what's going on. I'm doing that now. So far looks totally fine though.

jameshcorbett (approved):
I'm in!

jameshcorbett (changes requested):
Actually I think this is breaking `t2001-getrabbit.t 8 - flux rabbitmapping works in nested instances`. Perhaps related to https://github.com/flux-framework/flux-sched/pull/1380#issuecomment-3164435249.

grondo:
Is the test perhaps picking up the system Fluxion rc1 scripts and trying to load the module twice?
If modprobe is active then Fluxion modules should not be loaded by `flux module load`.

If this is reproducible I can take a look. It is important to get this right before moving forward.

grondo:
Oh I guess maybe one problem here may be that until https://github.com/flux-framework/flux-sched/pull/1380 is merged,
if the tests here run with `FLUX_RC_USE_MODPROBE` then flux-core is using modprobe, but flux-sched is not, so Fluxion modules will fail to load.

Perhaps the answer is to just wait until flux-core and Fluxion are sorted with support for the new moprobe scheme, then merge this PR (or re-review it). As noted above, the old rc1.d/* scripts will still work after moving to modprobe, so there's perhaps no real urgency.

jameshcorbett (approved):
https://github.com/flux-framework/flux-sched/pull/1380 is merged and the failing test passes now so I'd say set MWP!

garlick:
Any reason not to set MWP at this point @grondo or @jamescorbett?

jameshcorbett:
Go for it!

PR #385, opened 2025-08-11
dws: allow jobs to tolerate lost nodes and/or rabbits

### LINKED ISSUE #364: Allow rabbit jobs to tolerate loss of some rabbits and compute nodes during prolog
@behlendorf recently found that with large rabbit jobs, he was frequently finding that one or two rabbits would fail to create their file systems, or one or two compute nodes would fail to mount. This was a persistent problem, even with all the recent changes to mark nodes as `badrabbit` if they fail to mount. Perhaps this is because not many other rabbit jobs had been submitted, which might have helped filter out some of the bad nodes, or perhaps at that scale it simply becomes likely that something will fail; in any case, it would be useful if the rabbit prolog could proceed in spite of some user-specified number of failures. For example, a user might specify that they want 5050 nodes, and can tolerate the loss of 50 of them, or something similar.

In our discussion, only XFS and GFS2 could tolerate the loss of rabbits. With Lustre things would be more complicated and the consensus was that it would not be worth trying to make it work. However, I believe Lustre could tolerate the loss of compute nodes just fine.

jameshcorbett:
As described in #364, some rabbit jobs would benefit from the ability to tolerate lost nodes.

Fixes #364.

### DIFF
--- a/doc/guide/rabbit.rst
+++ b/doc/guide/rabbit.rst
@@ -143,6 +143,28 @@ in a batch script:
 	echo "Hello World!" > $DW_JOB_lustreproject/world.txt
 
 
+Enabling Rabbit Fault Tolerance
+-------------------------------
+
+Imagine you submit a ten-thousand-node Flux job that requests a rabbit file
+system. The job sits in the queue for a long time while your job is scheduled.
+Finally your job is assigned resources, but then a single node fails to mount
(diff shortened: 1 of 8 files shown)


jameshcorbett:
Thanks @cmoussa1, setting MWP!

PR #387, opened 2025-08-12
dws: add rc1 script to drain nodes with rabbit failures

jameshcorbett:
PR #385 added all the infrastructure to identify nodes with rabbit failures and nevertheless allow jobs to proceed through them. However, there still needs to be some action taken on the nodes with failures. This PR adds a rc1 script that drains nodes. In the future, it might be better to move the nodes to a separate queue, so that they are still available for use. However, that will be left for a later release.

Related to #364.

### DIFF
--- a/doc/guide/rabbit.rst
+++ b/doc/guide/rabbit.rst
@@ -164,6 +164,10 @@ tolerate the failed creation of file systems. If any rabbits fail to create thei
 Lustre targets, the whole job will fail. However, ephemeral Lustre rabbit jobs can
 still tolerate failed mounts.
 
+If this attribute is set, and a job proceeds through some rabbit failures, the nodes
+that are missing file systems will be drained when the allocation is granted. The
+nodes may still be undrained and used at the user's discretion.
+
 
 Fetching Rabbit Information
 ---------------------------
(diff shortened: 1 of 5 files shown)


doc/guide/rabbit.rst
```diff
@@ -164,6 +164,10 @@ tolerate the failed creation of file systems. If any rabbits fail to create thei
 Lustre targets, the whole job will fail. However, ephemeral Lustre rabbit jobs can
 still tolerate failed mounts.
 
+If this attribute is set, and a job proceeds through some rabbit failures, the nodes
+that are missing file systems will be drained when the allocation is granted. The
+nodes may still be undrained and used, at the user's discretion.
```
cmoussa1, line 169: minor nit: I _think_ you could remove this comma on the last line as I believe it causes a sentence fragment.

```diff
- nodes may still be undrained and used, at the user's discretion.
+ nodes may still be undrained and used at the user's discretion.
```

cmoussa1 (approved):
LGTM! One tiny documentation nit:

doc/guide/rabbit.rst
jameshcorbett: Hmmm yeah I think I like it better without the comma! Good catch, thanks. I removed it and force-pushed, setting MWP.

PR #391, opened 2025-08-20
testsuite: don't load deprecated barrier module

garlick:
Problem: the flux-core barrier module is deprecated, but is loaded in test.

Don't load it.  It is unused by flux-coral2.


chu11 (approved):
LGTM!

garlick:
Thanks - I'll set MWP.

PR #392, opened 2025-08-26
dws: stop prerun timer once state completes

jameshcorbett:
Problem: the timer for the `PreRun` state cancels a job if it gets stuck in that state for too long. However, the timer is not cleared when the state completes successfully, so jobs can be canceled improperly if they run for any considerable length of time.

Stop the timer when `PreRun` state completes.


cmoussa1 (approved):
LGTM!

PR #393, opened 2025-09-02
add cray-slingshot jobtap plugin for VNI reservation

garlick:
This is the first chunk of work to get VNI tagging working for the slingshot interconnect.  It is a jobtap plugin that reserves VNIs for each job in the system instance when it enters RUN state, posting an event to the job eventlog that can be accessed by other components yet to be added.

The plugin parses a config table that looks like this (defaults commented out):
```toml
[cray-slingshot]
#vni-pool = "1024-65535"
#vnis-per-job = 1
#vni-reserve-fatal = true
```

This uses a newer idset feature, IDSET_FLAG_ALLOC_RR, so it bumps the minimum required flux-core version to 0.76.  That feature enables  "round robin" allocation of  VNIs, which delays their reuse as long as possible, to make accidental reuse less likely in the event that one remains in use in a stuck process after the exec system gives up it and the job transitions to INACTIVE.

This particular component does not require the Cray  `libcxi` but it does use it if found to get the maximum number of VNIs per CXI service.  If `libcxi` is not found, a default is used, which makes this easier to test in CI.   Since future slingshot components will require `libcxi` to discover NIC hardware, it may be good to begin building RPMs with  `shs-libcxi-dev` added to the BuildRequires beginning with the first release that contains this plugin.

The general design for slingshot support was written up (though pretty raw!) [here](https://flux-framework.readthedocs.io/projects/flux-coral2/en/latest/guide/slingshot.html)


jameshcorbett (approved):
Looks great! One question.

garlick:
Thanks!  I'll set MWP then.

garlick:
rebased manually on current master

PR #396, opened 2025-09-02
add cray-slingshot shell plugin for VNI/CXI environment setup

garlick:
This adds a shell plugin that manages the slingshot VNI/CXI environment for jobs.

When running in a sub-instance (of Flux or of Slurm) where the broker has SLINGSHOT_* variables set based on a reservation, this ensures that they are inherited from the local broker and not set via the job environment (since they need to be different on each node).

When running in the system instance, this works in conjunction with the jobtap plugin of #393 to translate a reservation into environment settings.  To be effective it also needs help from a prolog and epilog/housekeeping to perform the privileged operation of creating and destroying CXI services to go with the reserved VNIs.  Those will come in a future PR.

Currently this is based on top of #393 (this PR is the last 4 commits)

Marking WIP for the moment because I need to circle back and try it on the real system again after making small changes to code that is not compiled if libcxi is not installed.

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -31,6 +31,7 @@ TESTSCRIPTS = \
 	t2001-getrabbit.t \
 	t2002-dws2jgf2.t \
 	t3000-slingshot-jobtap.t \
+	t3001-slingshot-shell.t \
 	python/t0001-directive-breakdown.py \
 	python/t0002-storage.py \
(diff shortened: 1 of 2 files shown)


garlick:
Rebased on master and dropped the WIP after testing on hetchy and making a few minor changes.

jameshcorbett (approved):
LGTM! Two little things

PR #397, opened 2025-09-03
add flux-slingshot utility

garlick:
Problem: prolog/epilog/housekeeping have to create/destroy CXI services as root, but the available tools, e.g.  cxi_service(1) will not be the most convenient to drive from a shell script.
    
Add a helper program that can also be built out as a Flux oriented Slingshot diagnostic if needed.

This is based on top of #396

I'll mark it WIP for now pending some manual testing.

### DIFF
--- a/etc/Makefile.am
+++ b/etc/Makefile.am
@@ -3,6 +3,13 @@ systemdsystemunit_DATA = flux-coral2-dws.service
 #endif
 
 fluxconfdir = $(sysconfdir)/flux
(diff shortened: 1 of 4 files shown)


garlick:
oops, fixed a formatting error

garlick:
Bah, my testing has been limited but I think this at least passes the hah hah.  Removing the WIP

garlick:
Thanks!  MPW now.

PR #403, opened 2025-09-04
shell: use eventlog_wait_for in dws_environment.c

jameshcorbett:
Problem: `dws_environment.c` waits for an event in a job's eventlog, but there is a utility function for doing so.

Use the common utility function.


jameshcorbett:
Oh, right. I think that should be OK though. I'll set MWP.

PR #408, opened 2025-09-13
fix slingshot CXI service resource request and improve error messages

garlick:
Problem: the slingshot prolog helper always fails to allocate a CXI service

Adjust max resource values in the allocation request so that they never exceed device specific hardware limits.

Tested on hetchy.


garlick:
Thank you!

PR #413, opened 2025-09-15
build: just warn if kubernetes is not found

garlick:
Problem: the kubernetes python module on TOSS 4 systems is too old to work with flux-coral2, which makes it a bit annoying to develop other aspects of flux-coral2 on TOSS.

Change the error to a warning so flux-coral2 can be configured on TOSS 4.

p.s. I'm not really sure if this is a good idea.  I defer to @jameshcorbett if there is a better way to deal with this.  I just keep adding this commit in my local development branches, which is a bit annoying to have to keep remembering.


garlick:
NP!  Thanks! I'll just use `PYTHON=/usr/bin/python3.6`.  No need to make this complicated.

Issue #412, opened 2025-09-15
slingshot: set SLINGSHOT_TCS environment variable


garlick:
Problem: [SLINGSHOT_TCS](https://flux-framework.readthedocs.io/projects/flux-coral2/en/latest/guide/slingshot.html#envvar-SLINGSHOT_TCS) is supposed to be set by the `cray-slingshot` shell plugin, but currently isn't.

garlick:
The word from HPE is
- this is used by Cray MPICH.  If unset, MPICH will use a default
- it should be set to a single hex value, e.g. `SLINGSHOT_TCS=0x0a`
- this is a bitmask of DEDICATED_ACCESS=0x1, LOW_LATENCY=0x2, BULK_DATA=0x4, BEST_EFFORT=0x8
- suggestion: the WLM could assign LOW_LATENCY|BEST_EFFORT by default, and add DEDICATED_ACCESS for high priority jobs

PR #418, opened 2025-09-16
set SLINGSHOT_TCS in job environment

garlick:
Problem: the cray-slingshot shell plugin does not set SLINGSHOT_TCS in the job environment, but Cray MPICH does use it if set.

Set it as discussed in #412


jameshcorbett (approved):
Looks reasonable, it doesn't look like the suggestion of
> suggestion: the WLM could assign LOW_LATENCY|BEST_EFFORT by default, and add DEDICATED_ACCESS for high priority jobs

is implemented though?

PR #421, opened 2025-09-16
flux-slingshot clean: fall back to default pool config

garlick:
Problem: `flux slingshot clean` doesn't know that there is now a default vni pool config.

Fix that.

Note that this isn't even used yet because it's a little scary.  At this point the housekeeping script just runs `flux slingshot epilog` with a longer `--retry-busy` duration than the epilog script.  IOW it only removes CXI services for the job that triggered housekeeping.  `clean` would remove _all_ CXI services matching Flux's VNI pool on the assumption that the node is exclusively scheduled and none should be active at that point.


jameshcorbett (approved):
Nice!

PR #428, opened 2025-10-14
dws: directivebreakdown chassis support

### LINKED ISSUE #401: Allow users to specify a number of racks (i.e. chassis) in jobspec resources
For node-local rabbit storage, some users may want to force Fluxion to pick a specific layout for their rabbit jobs, making sure all nodes are on a single chassis, or perhaps that each node is on its own chassis.

There will be a number of components to this:

- [x] Add a command-line plugin to add something like a `chassis` argument
- [ ] Allow jobspecs with this new resource type to pass validation steps
- [ ] Allow the `flux-coral2-dws` service to handle such jobspecs appropriately.

jameshcorbett:
Problem: code in `directivebreakdown.py` rejects jobs with top-level
'chassis' entries in jobspec. However, it would be useful to combine
rabbit allocations with the `--coral2-chassis` argument from the CLI
plugin introduced in #427.

Add support for chassis jobspecs in the `directivebreakdown` module.

Fixes #401.

### DIFF
--- a/t/t1002-dws-workflow-obj.t
+++ b/t/t1002-dws-workflow-obj.t
@@ -318,15 +318,23 @@ test_expect_success 'job submission with multiple valid DW strings on the same l
 	job_epilog_start_finish_clean $jobid
 '
 
-test_expect_success 'job submission with multiple valid DW strings in a JSON file works' '
+test_expect_success 'job submission with multiple valid DW strings in a JSON file works, with chassis' '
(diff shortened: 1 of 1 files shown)


wihobbs (approved):

jameshcorbett:
Thanks! I pushed a commit adding another test. Since it's a small commit and just tests, I'll set MWP.

PR #430, opened 2025-10-15
doc: add man page for flux run --coral2-chassis

jameshcorbett:
Problem: there is no documentation for the `--coral2-chassis` flag.

Add documentation.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -13,7 +13,8 @@ MAN1_FILES_PRIMARY = \
 	man1/flux-rabbitmapping.1 \
 	man1/flux-getrabbit.1 \
 	man1/flux-dws2jgf.1 \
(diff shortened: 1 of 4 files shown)


grondo:
A bit off topic, but in general we need a standardize way to document these cli plugins which could be provided by site config management, packaging, maybe eventually even per-user plugins.

Originally I had thought that the full documentation should somehow be supplied in the python docstring for the plugin, and some kind of `help` option could display that docstring. This would keep the docs with the plugin source and somewhat standardize its discovery by users. I'm not sure if that's sufficient, but maybe we'll revisit it in flux-core.

jameshcorbett:
Putting documentation in the docstring makes sense to me. This plugin certainly does not require much documentation, what it does is reasonably straightforward.

PR #431, opened 2025-10-15
build: add flux-config-rabbit to Makefile.am

jameshcorbett:
Problem: the flux-config-rabbit(5) man page is not included in Makefile.am.

Add it.

The first commit from #430.


garlick (approved):
LGTM!

PR #433, opened 2025-10-21
drop unnecessary timeouts that may cause large jobs to fail

### LINKED ISSUE #432: cray-slingshot: waiting for eventlog: error reading job eventlog: Connection timed out
Problem: on a system with cray-slingshot-0.27.0-1, a 2K node job failed with the message in the title.

The `cray-slingshot.so` jobtap plugin was not loaded.

garlick:
Problem: large jobs occasionally fail with ETIMEDOUT on the cray-slingshot eventlog watch.
   
The three shell plugins all use a similar strategy to wait 30s for an eventlog entry OR  a surpassing event to appear in the job eventlog.   The 30s timeout has been observed to be too short on El Cap,

Since the surpassing event is guaranteed by RFC 21, the timeout isn't really helping.
Just remove it, in all three shell plugins.
    
Fixes #432

### DIFF
--- a/src/shell/plugins/dws_environment.c
+++ b/src/shell/plugins/dws_environment.c
@@ -59,7 +59,7 @@ static int read_future (flux_shell_t *shell, flux_future_t *fut)
     json_t *context, *env;
     flux_error_t error;
 
-    if (eventlog_wait_for (fut, "dws_environment", 30.0, &context, &error) < 0) {
+    if (eventlog_wait_for (fut, "dws_environment", -1, &context, &error) < 0) {
         shell_log_error ("waiting for eventlog: %s", error.text);
         return -1;
     }


src/shell/plugins/cray-slingshot.c
```diff
@@ -53,6 +53,7 @@
 #endif
 
 #include "src/common/libutil/eventlog.h"
+#include "src/common/libutil/fsd.h"
```
jameshcorbett: how come?

jameshcorbett (approved):
This seems fine to me but do you think it's better than setting the timeout to something like 600?

garlick:
> This seems fine to me but do you think it's better than setting the timeout to something like 600?

Thanks for asking because the explanation I was giving myself for why these were unnecessary was a bit thin.  Looking deeper, here's what should happen if for some reason the eventlog watch blocks for a long time:

- dws and pals  would block in the `shell.init` callback.  This would cause the other shells to be stuck in the "init" barrier.
- slingshot would block in the `shell.post-init` callback.  This would cause the other shells to be stuck in the "start" barrier,

Those barriers have a timeout, by default 30m - see `barrier-timeout` in [flux-config-exec(5)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man5/flux-config-exec.html).  If this timeout is reached, the nodes that didn't enter the barrier are drained and the rest are returned to the scheduler.

So this contingency was planned for and we don't need further timeouts to ensure that resources are released in the event of a hang.

jameshcorbett (approved):
I'm in!

PR #434, opened 2025-10-22
dws: update dws and nnf crd versions

jameshcorbett:
Problem: the latest API versions for DWS and NNF are a release above what is used in the code base.

Increment the version strings, so that they are at the latest.

WIP until the versions referenced in the PR are deployed across LC.


cmoussa1 (approved):
LGTM!

Issue #337, opened 2025-04-03
dws and port_distributor jobtap plugins may crash broker when unloaded


jameshcorbett:
In [a comment](https://github.com/flux-framework/flux-coral2/pull/336#discussion_r2025817545) on #336 , @grondo pointed out that any dynamically-loaded plugin that does asynchronous work may cause the broker to crash:

> If the plugin is unloaded while the watcher is still active, this could cause a broker crash since the epilog_timeout_cb symbol won't exist anymore.
> 
> Unfortunately, in these situations, the plugin needs to keep a list of outstanding watchers so it can stop them and/or destroy them on exit. i.e. you can't rely on using the job aux item for this purpose.

I asked

>   How could I trigger cleanup on plugin removal? By using flux_plugin_set_aux or something, with the destructor set to my cleanup function?

and @grondo said:

> Yeah, I think the approach would be to have a global context stored in the plugin aux cache and add a list of objects to that ctx that would need to be freed on plugin removal.
> 
> Actually looking at existing plugins in core, many of them suffer from this same issue 🤦. Feel free to add an issue to address this at some point in the future. For now we'll have to be careful reloading most plugins that do asynchronous work. Maybe we can add something to the API to make this more manageable and obvious.

Both the dws-jobtap and cray_pals_port_distributor plugins have this vulnerability.

PR #467, opened 2026-04-03
dws: improve malformed RPC error message

jameshcorbett:
Problem: when RPC protocol changes between flux-coral2 versions, and flux-coral2 is updated on a live system, the `dws-jobtap` plugin may be left at an older version that is incompatible with the `coral2_dws` module.

Add a more descriptive error message and suggest restarting Flux to reload both the `coral2_dws` module and the `dws-jobtap` plugin.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -691,6 +691,13 @@ static void resource_update_msg_cb (flux_t *h,
                          &constraint_to_add)
         < 0) {
         errmsg = "received malformed dws.resource-update RPC";
+        if (flux_msg_unpack (msg, "{s:I}", "id", &jobid) == 0) {
+            raise_job_exception (p, jobid, "exception", errmsg);
+        }
+        flux_log_error (h,
+                        "%s, coral2_dws module and dws_jobtap plugin versions"
+                        " may be out of sync, Flux may need to be restarted",
+                        errmsg);
         goto error;
     }
     if (errmsg) {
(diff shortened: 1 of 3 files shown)


garlick (commented):
Would there be a downside to suggesting to restart the service then reload the jobtap plugin?  Restarting flux is simpler to describe but a bit heavyweight to carry out...

jameshcorbett:
My concern is #337. The suggestion could be "kill any rabbit jobs and ensure no new rabbit jobs are submitted. Then, reload the jobtap plugin and start the coral2_dws service." If there are active rabbit jobs, the jobtap plugin may have async work out there, and the broker may segfault. So to be really safe I suggested a flux restart.

cmoussa1:
What about raising a job exception on those rabbit jobs with this descriptive error message? I am not sure if that is any better from the user or admin's perspective if the job has an exception raised on it, but would that prevent them from having to a) kill the jobs manually, or b) restart a Flux instance entirely? And instead you could just reload the plugin after all of the rabbit jobs have finished cleaning up? 🤷

jameshcorbett:
> What about raising a job exception on those rabbit jobs with this descriptive error message? I am not sure if that is any better from the user or admin's perspective if the job has an exception raised on it, but would that prevent them from having to a) kill the jobs manually, or b) restart a Flux instance entirely? And instead you could just reload the plugin after all of the rabbit jobs have finished cleaning up? 🤷

That would definitely be better for (a)! It wouldn't help with (b) though unfortunately, they would still need to restart the instance or at least reload the jobtap plugin manually. I pushed up some code to raise an exception.

cmoussa1 (approved):
This LGTM!

One consideration (if possible) is keeping track of plugin versions (maybe via configuration file) and checking the version during `plugin_init ()` - if the versions are out of date, you could output the same message notifying the admins that it is outdated and you need to reload it and/or restart Flux. This is probably out of scope and could be work for a future PR, but just something you could think about, if anything!

PR #468, opened 2026-04-03
dws: fixes for rabbits as computes

jameshcorbett:
Problem: running jobs that requested both rabbit file systems and rabbit cores revealed some errors.


cmoussa1 (approved):
LGTM!

PR #469, opened 2026-04-07
dws: enable faulthandler module

jameshcorbett:
Problem: when the `coral2_dws` module is killed by systemd due to a watchdog timeout, it would be good to see a stack trace.

Enable the faulthandler module to print a stack trace on SIGABRT.


cmoussa1 (approved):
LGTM!

PR #470, opened 2026-04-08
flux_k8s: compare resource versions in `watch`

jameshcorbett:


milroy (approved):
This change looks correct to me.

PR #475, opened 2026-04-17
doc:  improve DWS/rabbit internals design document

garlick:
Problem: RPC service names for the jobtap plugin and dws service are reversed in a couple places, and the RPCs are not fully described.

Fix the reversals and add payload descriptions to the RPC section.

Add an example of the rabbitmapping file.

Improve some RST formatting.


jameshcorbett (approved):
Thanks for catching this / sorry I didn't get it on the first PR!

garlick:
Thanks!  I'll set MWP.
(And no prob, I hit you with a lot of text in one go!)

PR #477, opened 2026-04-18
dws: start epilog on finish event

### LINKED ISSUE #472: Too many nodes drained on elcap with message "failed to unmount rabbit"
@kkier reported that 4k and then later 11k jobs were drained with the message. The message comes from one place, line 401 in /bin/coral2_dws.py. The function is pretty simple: ask kubernetes for all nodes in the job that do not report themselves as being in the Unmounted state, and drain them.

The function is called if a workflow has been in postrun for longer than the configured timeout, which is 15 minutes on elcap.

jameshcorbett:
@grondo asked in #472 

> Why is the dws epilog action started on entry to CLEANUP state instead of on the finish event?
> 
> As noted above, the job-manager.epilog doesn't start until the finish event. Would it make sense to start the dws epilog at the same time (along with the timer)? Before the finish event is posted, there are still IMP, job shell, and maybe user processes on one or more nodes of the job, so it may not make sense to do whatever occurs in the DWS epilog action.
> 
> A case to handle here though is when a prolog fails there will be no start event and thus no finish event. In the perilog plugin the job-manager.epilog is started immediately when the job-manager.prolog fails. The dws jobtap plugin could handle this case by starting its epilog action on entry to the CLEANUP state when there was no start event (using flux_jobtap_job_event_posted()).

He made a good point. 

Fixes #472.

### DIFF
--- a/src/job-manager/plugins/dws-jobtap.c
+++ b/src/job-manager/plugins/dws-jobtap.c
@@ -444,16 +444,90 @@ cleanup:
     return;
 }
 
-static int finish_cb (flux_plugin_t *p, const char *topic, flux_plugin_arg_t *args, void *arg)
+/* Helper function to start DWS epilog for a job.
+ * Called by both finish_cb (for jobs that ran) and cleanup_cb (for jobs that
+ * failed before starting). The dws_run_started parameter indicates whether the
+ * job actually started (1) or not (0), which is passed to the coral2_dws service.
+ */
+static int start_dws_epilog_for_job (flux_plugin_t *p, flux_jobid_t id, int dws_run_started)
 {
-    flux_jobid_t id;
-    json_t *dw = NULL;
     flux_future_t *post_run_fut;
     flux_t *h = flux_jobtap_get_flux (p);
     flux_reactor_t *r = flux_get_reactor (h);
     flux_watcher_t *watcher = NULL;
-    int dws_run_started = 0;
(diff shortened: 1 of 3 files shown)


src/job-manager/plugins/dws-jobtap.c
```diff
@@ -444,17 +444,91 @@ static void epilog_timeout_cb (flux_reactor_t *r, flux_watcher_t *w, int revents
...
     struct create_arg_t *create_args = NULL;
 
+    if (!r) {
+        flux_log_error (h, "Failed to fetch reactor from handle for %s", idf58 (id));
+        current_job_exception (p, "Failed to fetch reactor from handle");
+        return -1;
+    }
+    if (!(create_args = calloc (1, sizeof (struct create_arg_t)))
+        || flux_jobtap_job_aux_set (p, FLUX_JOBTAP_CURRENT_JOB, NULL, create_args, free) < 0) {
+        free (create_args);
```
garlick: free clobbers errno

garlick (approved):
This looks good to me, but no tests?

jameshcorbett:
> This looks good to me, but no tests?

Hmmm, well it passing the existing test suite is a lot, because the behavior is supposed to be the almost identical. The only time the new behavior would differ from the old would be if there were a stuck process or something, so the `finish` event were not posted for a while. But I added some tests regardless.

Issue #473, opened 2026-04-14
Marking rabbits with external allocations


jameshcorbett:
Some amount of rabbits are allocated externally by `systemstorage` resources. If Flux could watch the `servers` resource type (and ignore any servers that are associated with jobs), it could get updates about how much storage is being used by persistent file systems and system storages. However, it would then need to inform Fluxion about these allocations.

See also #21

trws:
IIRC, one way we discussed to handle this was to add a way to update the pool size with an RPC right @jameshcorbett?  There's an inherent race condition in it since there's always going to be a gap between the update and our finding out in the scheduler, but I'm not sure what we can do about that to be honest.  It's also closely related to a request we got for being able to dynamically update license counts from outside.

jameshcorbett:
> IIRC, one way we discussed to handle this was to add a way to update the pool size with an RPC right @jameshcorbett?

I don't remember discussing that recently but it sounds like it would do the trick. The race condition is probably tolerable.

Currently, the external allocations are fairly stable. After discussing with @behlendorf , we could minimize this issue in the short term by initializing Fluxion with a reduced SSD pool size, to account for the external allocations. But the RPC solution doesn't seem too difficult to implement so we should probably go with that.

trws:
I think it's mainly just something we haven't done, it should be straightforward as long as we keep it simple and make it an error to reduce the pool size below any active allocation size and things like that.

jameshcorbett:
Using the code from PR #479, I updated the `rabbitmapping` file on all OCF clusters to account for persistent storages like `/l/ssd`. So this problem is now mitigated.

trws:
Would you call this ready to close for now @jameshcorbett or is there something in flight you want to track here?

jameshcorbett:
I'm not sure we should close this, but maybe we should remove it from the sprint or somehow mark it as low priority. If users/admins ever start using lots of space in persistent storages, this issue would bite us again, and we'd want a way to update the vertex sizes dynamically.

trws:
Ok, we could move it to a resolved state on the board or similar I guess.
________________________________

PR #479, opened 2026-04-28
dws: external rabbit allocations in `flux rabbitmapping`

jameshcorbett:
Mitigates #473 by reducing rabbits' capacity in config. This is not the full solution however, because ideally it would be more dynamic, and Fluxion would be able to update capacity as new allocations are made on a live system.

### DIFF
--- a/src/cmd/flux-rabbitmapping.py
+++ b/src/cmd/flux-rabbitmapping.py
@@ -116,6 +116,11 @@ def main():
         action="store_false",
         help="Do not sort keys in output JSON document",
     )
+    parser.add_argument(
+        "--ignore-external-allocations",
+        action="store_true",
+        help="Ignore external (non-Flux) allocations from Servers objects",
+    )
     args = parser.parse_args()
     k8s_api = cleanup.get_k8s_api(args.kubeconfig)
     crd.determine_api_versions(flux.Flux(), k8s_api)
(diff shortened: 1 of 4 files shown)


garlick:
I think the sharness test needs to be added to the Makefile.am so it actually gets run?

jameshcorbett:
Oh duh good catch. I was running the test locally with `FLUX_TESTS_LOGFILE=t make check TESTS=t2003-rabbitmapping.t` but I guess that somehow bypasses that the test isn't in the makefile?

garlick:
Spot checking the builders, it looks like maybe the rabbitmapping test is being skipped?  At least the first four that I checked reported e.g.
```
2026-04-28T21:07:54.9403286Z      t2003-rabbitmapping.t:  SKIP: N=1   PASS=0   FAIL=0 SKIP=1 XPASS=0 XFAIL=0
```

jameshcorbett:
> Spot checking the builders, it looks like maybe the rabbitmapping test is being skipped? At least the first four that I checked reported e.g.
> 
> ```
> 2026-04-28T21:07:54.9403286Z      t2003-rabbitmapping.t:  SKIP: N=1   PASS=0   FAIL=0 SKIP=1 XPASS=0 XFAIL=0
> ```

That's expected, the test requires the dws kind environment, which isn't set up in CI. I'm verifying that the test suite passes locally, but it's still running...

jameshcorbett:
Can confirm, tests pass locally as part of a regular `make check`, i.e. without `TESTS=...`

garlick (approved):
LGTM based on a light review.  I opened an issue to capture the CI deficiency (I realize it was well known - sorry I forgot :-)
- #480

PR #485, opened 2026-05-19
Rabbit frobnicator

jameshcorbett:
Problem: Jobs with DW directives need to be able to run on rabbit
nodes that have the 'rabbit' property but are not configured for any
queue.

Add a frobnicator plugin that extends jobspecs with DW directives and
'allow_rabbits=true' attribute to include an OR constraint combining
existing constraints with the 'rabbit' property. This allows jobs to
match either their queue's compute nodes or any rabbit node.

The plugin creates a top-level OR constraint that wraps any existing
constraints from the constraints frobnicator with a rabbit property
constraint, enabling flexible scheduling across both resource types.

### DIFF
--- a/src/python/flux/job/frobnicator/plugins/rabbit.py
+++ b/src/python/flux/job/frobnicator/plugins/rabbit.py
@@ -1,0 +1,54 @@
+##############################################################
+# Copyright 2026 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+##############################################################
+
+"""Add rabbit property constraint to jobspec with DW directives.
+
+For jobs with both a 'dw' attribute and 'allow_rabbits' set to true,
+this plugin adds a top-level OR constraint that combines any existing
+constraints with the 'rabbit' property. This allows jobs to match
+either their existing constraints or nodes with the 'rabbit' property.
+
+This plugin is meant for systems where rabbit nodes have the `rabbit`
+property and are not configured for any queue.
+"""
+
+from flux.job.frobnicator import FrobnicatorPlugin
+
+
+class Frobnicator(FrobnicatorPlugin):
+    """Frobnicator plugin to add rabbit property constraints for DW jobs"""
+
+    def frob(self, jobspec, *_args, **_kwargs):
+        """Add OR constraint with rabbit property if job uses DW and allows rabbits
+
+        Args:
+            jobspec: The jobspec to modify
+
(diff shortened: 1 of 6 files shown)


src/python/flux/job/frobnicator/plugins/rabbit.py
```diff
@@ -0,0 +1,51 @@
...
+            "system", {}
+        ) or not jobspec.attributes.get("system", {}).get("allow_rabbits", False):
+            return
+        # both conditions are met, add OR constraint to existing constraints (if any)
+        try:
+            existing_constraint = jobspec.attributes["system"]["constraints"]
+        except KeyError:
+            # No existing constraints - this shouldn't happen because the constraints
+            # frobnicator should have added the queue property, so do nothing
+            return
```
grondo: Should an error be raised here instead of silently doing nothing? This would make the condition easier to diagnose if a job somehow got into a state where this plugin was called without any constraints yet set on a job.

grondo (approved):
LGTM! One suggestion inline.

I also wonder if it is feasible to add an end-to-end integration test to this PR: e.g. set up a test instance with queues, some test broker ranks with a rabbit attribute, configure the frobnicator and make sure the job runs on the "rabbit" nodes.

src/python/flux/job/frobnicator/plugins/rabbit.py
jameshcorbett: Are there no odd cases where a job might not have constraints set? Like somehow an instance owner can submit a job that skips the `constraints` plugin? I'm concerned about potential consequences I guess.

jameshcorbett:
> LGTM! One suggestion inline.
> 
> I also wonder if it is feasible to add an end-to-end integration test to this PR: e.g. set up a test instance with queues, some test broker ranks with a rabbit attribute, configure the frobnicator and make sure the job runs on the "rabbit" nodes.

That could be done, but for a real integration test the "rabbit" nodes would also need to be given the special "storage_node" type to Fluxion, which would take some work and maybe some additional code changes (i.e. not just more test code). I'm going to punt on that for now!

src/python/flux/job/frobnicator/plugins/rabbit.py
grondo: The question is if the instance owner would bypass the queue constraints frobnicator plugin, but then also set the `dw` or `allow_rabbits` attribute in their jobspec, and expect their job to silently run with no constraints? I'd expect an error, but If that's acceptable then this is fine. 

I could be misreading the plugin here, but if a job was submitted with no constraints and no `dw` or `allow_rabbits` attr, then this plugin is skipped?

jameshcorbett: Good point! I changed it to raise an error and updated the tests. Setting MWP.

PR #486, opened 2026-05-20
etc: remove NNF clientmount prolog/epilog scripts

### LINKED ISSUE #478: can nnf-clientmount  be allowed to run continuously?
Problem: early in CORAL2 integeration, `nnf-clientmount` was identified as a computational noise source source, and steps were taken to keep it from interfering with jobs.  Specifically, we stop it in the prolog and start it in the epilog.  However, this has clouded thinking about timing and synchronization of flux with nnf workflows somewhat, and we wonder if it is still necessary.

Currently the `zstop_nnf_clientmount.sh` (a prolog snippet provided by flux-coral2) runs
```sh
#!/bin/bash

# stop the nnf-clientmount service.
# if the job's jobspec has a .attributes.system.dw entry, it uses
# rabbits and so the clientmount service should not be stopped until
# the mount has completed, which is indicated by the
# `dws_environment` event being posted to the job's eventlog.
# This script should run last because the `flux job wait-event` may
# take some time waiting, hence the 'z' prefix in the filename.

flux getrabbit || exit 0  # exit if not configured with rabbits

if flux job info ${FLUX_JOB_ID} jobspec | jq -e .attributes.system.dw > /dev/null ; then
    # the nnf-clientmount service should already be running. However, if
    # it isn't running, the job could hang while waiting
    # for the mounts. Just in case, issue a `systemctl start`.
    systemctl start nnf-clientmount
    TIMEOUT=$(flux config get --default=1800 rabbit.prolog_timeout)
    flux job wait-event -t${TIMEOUT} ${FLUX_JOB_ID} dws_environment || \
        flux job raise -t prolog-timeout ${FLUX_JOB_ID} "dws_environment timeout"
    systemctl stop nnf-clientmount
else
    systemctl stop nnf-clientmount
fi
```

and the `start_nnf_clientmount.sh` script (epilog script provided by flxu-coral2) runs

```ssh
#!/bin/bash

flux getrabbit || exit 0  # exit if not configured with rabbits

# start the nnf-clientmount service
systemctl start nnf-clientmount
```

We would like to revisit this and determine if `nnf-clientmount` can be allowed to run continuously if 
1) contained within the reserved "system cores" (It should be now.  This practice had not been adopted when the problem was originally discovered).
2) if that is not sufficient, using a systemd override to further constrain its cpu or network utilization

If it is determined that stopping it is unnecessary, then these prolog/epilog snippets can be dropped.

@kkier volunteered to find the noise benchmark used by HPE to identify this originally.

@wihobbs volunteered to try running the benchmark on a single node with and without `nnf-clientmount` running to gauge the effects.

Also ping @behlendorf, @jameshcorbett, and @trws

jameshcorbett:
Problem: the start_nnf_clientmount.sh and zstop_nnf_clientmount.sh prolog/epilog scripts are no longer needed for NNF storage management.

Remove start_nnf_clientmount.sh and zstop_nnf_clientmount.sh from etc/ and remove their references from etc/Makefile.am.

Fixes #478. Waiting for @wihobbs to give the approval after some further testing.


garlick (approved):
Sounds good to me!

wihobbs:
If we get rid of this here, we need to add it back in ansible (note my comment in #478)

jameshcorbett:
> If we get rid of this here, we need to add it back in ansible (note my comment in #478)

Better to keep it in the RPM. I'll close this PR, we can always reopen it if something changes.

PR #488, opened 2026-06-03
flux_k8s: use bulk set_ and remove_property RPCs

jameshcorbett:
Problem: flux-coral2 can send a huge number of set_ and remove_property RPCs, but Fluxion supports bulk operations.

Batch up the RPCs.


wihobbs (approved):
mrow!

PR #491, opened 2026-07-01
dws2jgf: do not output any ranks in `resource.exclude`

### LINKED ISSUE #490: flux-dws2jgf: remove `resource.exclude` ranks from JGF
Problem: there is no reason to include the `resource.exclude` ranks in JGF, since such ranks can never be used by Fluxion. They only bloat the graph and may also exacerbate https://github.com/flux-framework/flux-sched/issues/1490

jameshcorbett:
Problem: there is no reason to include the resource.exclude ranks in JGF, since such ranks can never be used by Fluxion. They only bloat the graph and may also exacerbate https://github.com/flux-framework/flux-sched/issues/1490.

Do not output any vertices for nodes that are listed in `resource.exclude`.

Fixes #490 .

### DIFF
--- a/src/cmd/flux-dws2jgf.py
+++ b/src/cmd/flux-dws2jgf.py
@@ -47,10 +47,21 @@ class Coral2Graph(FluxionResourceGraphV1):
     CORAL2 Graph:  extend jsongraph's Graph class
     """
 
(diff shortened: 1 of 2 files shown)


cmoussa1 (approved):
LGTM! Nice improvement. You might consider adding a brief description to this PR before merging, though!

PR #494, opened 2026-07-29
dws: remove `--disable-fluxion` logic

### LINKED ISSUE #484: Remove all `--disable-fluxion` logic in `coral2_dws.py`
Problem: there is a bunch of logic in `coral2_dws.py` built up to work around Fluxion not being able to schedule the rabbits. However, on all of the rabbit clusters, Fluxion is scheduling the rabbits, so that code is no longer needed. Once that stabilizes and we're sure we won't be reverting, we can remove all of that extra code.

jameshcorbett:
Problem: the coral2_dws service accepts a --disable-fluxion flag that
selects a non-Fluxion rabbit scheduling path. Fluxion now always
schedules rabbits, so the flag and its code paths are dead.

Fixes #484.


src/python/flux_k8s/storage.py
```diff
@@ -95,6 +92,34 @@ def rabbit_state_change_cb(self, event):
...
+        name = rabbit["metadata"]["name"]
+        if name not in self._rabbit_rpaths:
+            LOGGER.error(
+                "Encountered an unknown Storage object '%s' in the event stream", name
+            )
+            return
+        status = _get_status(rabbit, "Down")
+        self._mark_rabbit(status, name)
+        # TODO: add some check for whether rabbit capacity has changed
+        # TODO: update capacity of rabbit in resource graph (mark some slices down?)
```
cmoussa1, line 104: Are these `#TODO` items left on purpose? No problem if they are.

cmoussa1 (approved):
LGTM! One minor question

src/python/flux_k8s/storage.py
jameshcorbett: They're existing, so I think out of scope for this PR, and I think they shouldn't be removed yet anyway, since the TODO still (kind of) remains to be done.

PR #497, opened 2026-08-19
`flux rabbitmapping`: do not double-count allocations

### LINKED ISSUE #496: flux-rabbitmapping double-counts allocations
`flux rabbitmapping` takes the `.status.capacity` field of `storage` resources and then subtracts `systemstorage` capacity from it to get a baseline free capacity. However, as described in [this issue](https://github.com/DataWorkflowServices/dws/issues/284), `.status.capacity` is actually a measure of free capacity, not total capacity. So Flux double-counts allocations, reducing the capacity that should be considered available to jobs.

jameshcorbett:
Problem: populate_from_storages set each rabbit's capacity from Storage
status.capacity, which is free (not total) bytes. Subtracting external
Servers allocations from a free-space figure double-counts them; on
tuolumne209 this reported ~21.69 TiB free instead of ~24.82 TiB.
    
Sum status.devices[].capacity to get the rabbit's total raw capacity, so
that reduce_capacity_by_servers subtracts external allocations from the
true total rather than from an already-net free figure.

Fixes #496.

### DIFF
--- a/src/cmd/flux-rabbitmapping.py
+++ b/src/cmd/flux-rabbitmapping.py
@@ -42,7 +42,12 @@ def populate_from_storages(storages, rabbit_mapping):
                     f"connected to {compute_name} but systemconfiguration gives "
                     f"{rabbit_mapping['computes'].get(compute_name)}"
                 )
-        capacity = nnf.get("status", {}).get("capacity", 0)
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM!

PR #498, opened 2026-08-19
docs: document the `--coral2-rabbit-cores` option

jameshcorbett:
Problem: the experimental --coral2-rabbit-cores option is undocumented
in the rabbit guide.

Add a "Rabbit Cores" section to the Node Distribution part of the rabbit
guide describing the flag and its dependency on --coral2-chassis.


cmoussa1 (approved):
niceeee

PR #501, opened 2026-09-03
rabbitmapping: capacity check

jameshcorbett:
Problem: after reducing rabbit capacity by external (non-Flux)
allocations, nothing checks the result against the Storage resource's
reported free capacity. Since status.capacity already excludes external
allocations, total minus external should never fall below it.

Capture each Storage's status.capacity and compare it with the adjusted
capacity. If the values are inconsistent, print a warning to stderr
while preserving the JSON mapping on stdout and allowing the command to
finish.


cmoussa1 (approved):
LGTM!

